PR #4, opened 2017-06-13
Allow passing `Result<T, E>` to Lua

### LINKED ISSUE #3: Support passing Results to Lua
We already have a `ToLua` impl for `Option`, which will push the contained value if it exists, and `nil` otherwise. This maps pretty well into Lua's world.

Another common Lua pattern occurs when dealing with fallible operations like `io.open`, which either returns the opened file on success, or `nil` followed by an error message on failure. It would be nice if I could map Rust's `Result` to this idiom. This would require a `ToLuaMulti` impl instead of a `ToLua` impl, though (since the operation can push `nil` *and* the error message).

jonas-schievink:
For this to work, both `T` and `E` need to implement `ToLua`. An `Ok(t)`
passes the contained `T`, while an `Err(e)` passes `nil` followed by the
contained `E`. This matches the common Lua idiom used by functions like
`io.open`.

Closes #3

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -15,6 +15,22 @@ impl<'lua> FromLuaMulti<'lua> for () {
     }
 }
 
+impl<'lua, T: ToLua<'lua>, E: ToLua<'lua>> ToLuaMulti<'lua> for Result<T, E> {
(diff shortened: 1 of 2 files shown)


kyren:
Yeah, that makes complete sense to me.  Excellent work!

PR #5, opened 2017-06-13
Remove src/macros.rs

jonas-schievink:
It's not used by anything. It's not even included via `mod macros;`.


kyren:
Yep, complete oversight on my part.  Forgot to remove it when it was moved to a separate crate.

PR #6, opened 2017-06-13
Utilize `cfg_attr` instead of duplicating code

jonas-schievink:
Pretty minor nit I found...


kyren:
Huh, I didn't even know you could do that, thanks a lot!

PR #7, opened 2017-06-16
Enable `io.popen` on Linux

jonas-schievink:
Also adds a REPL example so I could test this :)

Basically, this is a port of https://github.com/tomaka/hlua/pull/132.

Now, there's a few more things we could do. `LUA_USE_LINUX` enables a bit more than just `io.popen`, and it should work fine on a few other OSes, but I didn't bother checking where exactly it is supported.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,17 @@
 extern crate gcc;
 
+use std::env;
+
 fn main() {
-    gcc::Config::new()
-        .define("LUA_USE_APICHECK", None)
+    let mut config = gcc::Config::new();
+
+    if env::var("CARGO_CFG_TARGET_OS") == Ok("linux".to_string()) {
+        // Among other things, this enables `io.popen` support.
+        // Despite the name, we could enable this on more platforms.
+        config.define("LUA_USE_LINUX", None);
+    }
+
+    config.define("LUA_USE_APICHECK", None)
         .include("lua")
         .file("lua/lapi.c")
         .file("lua/lauxlib.c")
--- a/examples/repl.rs
+++ b/examples/repl.rs
@@ -1,0 +1,30 @@
+//! This example shows a simple read-evaluate-print-loop (REPL).
+
+extern crate rlua;
+
+use rlua::*;
+use std::io::prelude::*;
+use std::io::{stdin, stdout, stderr, BufReader};
+
+fn main() {
+    let lua = Lua::new();
+    let mut stdout = stdout();
+    let mut stdin = BufReader::new(stdin());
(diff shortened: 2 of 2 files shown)


kyren:
Yeah, not only linux, we really should be defining all of LUA_USE_LINUX, LUA_USE_WINDOWS, LUA_USE_MACOSX, LUA_USE_POSIX correctly at least.  I sort of punted on that when I was setting this up because I just didn't need it at the time.  I can't remember if there are any more realistic platform defines than that, I don't think so?

The repl example is a SUPER good idea, but you can do even better than that, there's an IncompleteStatement error specifically to make repls with, so you can print '>>' like the normal lua repl.

I'm going to merge your pull request soon, but I'm going to wait until I have a free moment to do the changes I mentioned right afterwards.  Internally at chucklefish we have gitlab-ci build servers for our projects on all pc architectures, so I should be able to do the define changes and at least mostly test it on macos / linux / windows without too much headache.

By the way, I'm still kind of a github scrub (gitscrub?), so I'm not too familiar with common workflows.  There's not a way to like.. collaborate on a pull request is there?  I don't know what the proper workflow here is.. I have a set of related changes to your pull request that I would like to make, and I can either 1) make the changes suggested by the pull request then close it, 2) merge it then immediately work on the changes in new commits, or 3) somehow collaborate on it?  It's not terribly important here, I push commits all the damn time anyway, I mostly just don't want to violate some github etiquette I'm not aware of.

By the way thanks for all the great pull requests! 👍

Timidger:
> By the way, I'm still kind of a github scrub (gitscrub?), so I'm not too familiar with common workflows. There's not a way to like.. collaborate on a pull request is there? I don't know what the proper workflow here is.. I have a set of related changes to your pull request that I would like to make, and I can either 1) make the changes suggested by the pull request then close it, 2) merge it then immediately work on the changes in new commits, or 3) somehow collaborate on it? It's not terribly important here, I push commits all the damn time anyway, I mostly just don't want to violate some github etiquette I'm not aware of.

Assuming @jonas-schievink didn't forbit it explicitly, you can push directly to the pr [see here for more details](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/)

jonas-schievink:
@kyren Yeah, you should be able to just push to the branch. I've never used that feature either, though.

kyren:
> Assuming @jonas-schievink didn't forbit it explicitly, you can push directly to the pr see here for more details

Thanks!  I might not do that this time but that's good to know.

Issue #8, opened 2017-06-18
"not enough elements in stack" assertion when playing with coroutines


jonas-schievink:
Hit this while writing docs and an example for `LuaThread::resume`. Take this code:

```rust
extern crate rlua;
use rlua::*;
                                                             
fn main() {
let lua = Lua::new();
let thread: LuaThread = lua.eval(r#"
    coroutine.create(function(arg)
        assert(arg == 42)
        local yieldarg = coroutine.yield(123)
        assert(yieldarg == 43)
        return 987
    end)
"#).unwrap();
                                                             
assert_eq!(thread.resume::<_, u32>(42).unwrap(), Some(123));
assert_eq!(thread.resume::<_, u32>(43).unwrap(), Some(987));
assert_eq!(thread.resume::<_, u32>(()).unwrap(), None);
}
```

Executing it hits this assertion inside Lua:
```
rust_out: lua/ldo.c:663: lua_resume: Assertion `(((L->status == 0) ? nargs + 1 : nargs) < (L->top - L->ci->func)) && "not enough elements in the stack"' failed.
```

kyren:
Yeah, crashes there don't surprise me, good catch!  I'll definitely try and fix that tonight.

kyren:
Fixed by 7b00ff0, it wasn't even a deep bug I just skipped the part where I actually checked the status of the coroutine before attempting to resume it :/

@jonas-schievink, what do you think about the thread API here, I copied the API for threads from starbound's c++ lua api (which returns a Maybe<R>), but maybe this doesn't map to rust very well?

I'm thinking about changing the API to have a CoroutineDead error and just returning that, that makes more sense to me.  What do you think?

jonas-schievink:
That's probably a good idea. APIs that nest option and result can get confusing fast.

PR #9, opened 2017-06-18
Documentation

jonas-schievink:
Went through `lua.rs` and documented most items a bit. Docs can be improved further (and examples should be added), but this is a start.

### DIFF
--- a/src/lua.rs
+++ b/src/lua.rs
@@ -15,33 +15,49 @@ use ffi;
 use error::*;
 use util::*;
 
-/// A rust-side handle to an internal Lua value.
+/// A dynamically typed Lua value.
 #[derive(Debug, Clone)]
 pub enum LuaValue<'lua> {
+    /// The Lua value `nil`.
     Nil,
+    /// The Lua value `true` or `false`.
     Boolean(bool),
+    /// A "light userdata" object, equivalent to a raw pointer.
     LightUserData(LightUserData),
+    /// An integer number.
+    ///
+    /// Any Lua number convertible to a `LuaInteger` will be represented as this variant.
     Integer(LuaInteger),
+    /// A floating point number.
     Number(LuaNumber),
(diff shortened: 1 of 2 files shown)


kyren:
This is really great work.  I gave everything a quick read through and it all seems really good to me.  Some thoughts:

Thanks for letting me know I'm missing metamethods for lua 5.3, I'm going to add those tonight as part of a semver bump with changes related to #8.

You documented Lua::pop_ref, which is an internal function.  What is the right thing to do there?  This is an honest question about rust style as much as anything else, it shouldn't have a doc comment right because it won't show up in public docs, or should I be documenting my private functions in that way as well?  It's a fine comment we should maybe document all the complicated private methods there I just want to know what the right style is.

You mentioned that excess LuaMultiValues should be ignored, which is definitely true and matches lua semantics when calling lua functions and assigning multiple values like you said.  There's a dual to this actually, which is really important when using this API:  Missing values can be assumed to all be Nil.  If you unwrap HList![i32, Option<i32>] for example on the multi-value hlist![3], you'll get hlist![3, None] because multi values kind of have infinite Nils after the final value, which is just the same semantics you mentioned but just the other way around.

I'm not sure how best to explain that though, or where the best place to do so is.  It's really important mostly for accepting optional function arguments in callbacks.

kyren:
I'm gonna go ahead and merge this though before making a 0.6, it's all really good stuff so there's no reason to wait for a few super minor nits and letting it diverge.

jonas-schievink:
> You documented Lua::pop_ref, which is an internal function. What is the right thing to do there?

It's not required to document internal functions, but it's not forbidden either. So it makes sense when the function is important or nontrivial. Of course, using `///` doc comments doesn't *do* anything there, but that's still what they're made for, so I always use doc comments instead of normal comments (might also be handy for IDE support, some show the doc comments in a tooltip).

PR #10, opened 2017-06-18
Rename a few methods to better match their intention or conventions

jonas-schievink:
The commits should be self-explanatory.

I'm still working on some other breaking changes, so you might want to hold off on releasing a new version for now.

### DIFF
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -47,7 +47,7 @@ fn examples() -> LuaResult<()> {
     array_table.set(1, "one")?;
     array_table.set(2, "two")?;
     array_table.set(3, "three")?;
-    assert_eq!(array_table.length()?, 3);
+    assert_eq!(array_table.len()?, 3);
 
     let map_table = lua.create_empty_table()?;
     map_table.set("one", 1)?;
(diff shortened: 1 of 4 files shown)


kyren:
I think I agree with all of them other than set -> insert, but I don't feel super strongly about it.  It doesn't return the old value like map insert, nor should it really.

jonas-schievink:
Yeah, should I remove that commit?

kyren:
Yeah I think so, unless you feel differently?

jonas-schievink:
Nah, I've only included that for consistency with Rust maps. Since the function doesn't return the old value anyways, it doesn't matter too much. I've force-pushed the commit away :)

jonas-schievink:
> I'm still working on some other breaking changes, so you might want to hold off on releasing a new version for now.

Actually, scratch that. What I wanted wasn't as simple as I hoped.

kyren:
> Actually, scratch that. What I wanted wasn't as simple as I hoped.

What was the change you wanted to make, something I can assist you with?

jonas-schievink:
I was confused about the array methods defined on `LuaTable` and would have expected them to act on the sequence part of the table instead of asserting that the table is array-like (a notion that Lua does not have), but the current behaviour makes sense, too.

My initial plan was to replace them by methods acting on that sequence part, effectively making `for_each_array_value` equivalent to `ipairs`, but looking at how the `ToLua` impls for `HashSet` and `BTreeSet` work, the current behaviour makes sense, too.

Actually, let me open an issue about this so we can discuss this further.

PR #12, opened 2017-06-21
Add `Lua::load` for precompiling chunks

jonas-schievink:
This allows loading scripts into memory, performing a syntax check and
precompiling them, without immediately running them like `exec` and
`eval` do.


kyren:
LGTM, I'm gonna merge this and then investigate writing exec and eval in terms of load, or at least simplifying them a bit.

PR #16, opened 2017-07-16
Give `Lua::eval` a source name param and simplify

jonas-schievink:


kyren:
Seems very sensible to me!

Issue #15, opened 2017-07-16
Consider removing the `Lua` prefix from types/traits


jonas-schievink:
The `rlua` crate is already a namespace for the contents, so they don't need a prefix for disambiguation. Removing the prefix would allow writing shorter and slightly clearer code in my opinion.

Downside: `use rlua::*;` imports too much (including a `Result` type alias shadowing Rust's own result type). This can be mitigated somewhat by introducing a `prelude` module like many libs do. It's also not clear to me if renaming the `LuaString` type to `String` is a good idea.

(also, `LightUserData` is the only type that's not prefixed)

jonas-schievink:
While we're at it, the trait I have to implement to make something a userdata object is named `LuaUserDataType`, which is really long. We could rename it to `UserData`, and the current `LuaUserData` could be `AnyUserData` or `UserDataAny` to make clear that it's doing runtime type checking like `std::any::Any`.

Timidger:
Personally, I think it's fine to name things somewhat generically in Rust since they are namespaced (e.g `String`) **however** you might need to do some refactoring in places (e.g a function that returns a `std::string::String` would not compile if `rlua::String` is in scope).

You shouldn't really be doing `use <lib>::*` except in extreme cases IMHO, it makes it much harder for readers to know where types come from (especially because IDE support for Rust is lacking right now) and it can introduce problems like this. So I wouldn't worry about the potential downside that this makes a glob import unusable.

kyren:
PR has been merged, so I'm going to mark this issue as fixed.  I added a 'prelude' module for glob imports, but I don't know whether it's "really" worth it or not, I guess it's something to consider before a hypothetical 1.0 release.  *I* will use the prelude module though, so it's worth it for me at least.

PR #17, opened 2017-07-18
Remove the `Lua*` prefix from most types

jonas-schievink:
cc #15

Doesn't touch `LuaString` mainly because that's a *lot* of renaming work
and the code looks weird. Also I want feedback before I proceed.

### DIFF
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -4,7 +4,7 @@ extern crate rlua;
 
 use std::f32;
 
-use rlua::*;
+use rlua::{Lua, Result, Function, Variadic, UserData, UserDataMethods, MetaMethod};
 
 fn examples() -> Result<()> {
     // Create a Lua context with Lua::new().  Eventually, this will allow
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -6,7 +6,7 @@ use std::f32;
 
 use rlua::*;
 
-fn examples() -> LuaResult<()> {
+fn examples() -> Result<()> {
     // Create a Lua context with Lua::new().  Eventually, this will allow
     // further control on the lua std library, and will specifically allow
     // limiting Lua to a subset of "safe" functionality.
@@ -56,7 +56,7 @@ fn examples() -> LuaResult<()> {
     let v: i64 = map_table.get("two")?;
     assert_eq!(v, 2);
 
-    // You can pass values like LuaTable back into Lua
+    // You can pass values like Table back into Lua
 
     globals.set("array_table", array_table)?;
     globals.set("map_table", map_table)?;
@@ -76,7 +76,7 @@ fn examples() -> LuaResult<()> {
 
     // You can load lua functions
 
-    let print: LuaFunction = globals.get("print")?;
+    let print: Function = globals.get("print")?;
     print.call::<_, ()>("hello from rust")?;
 
     // This API handles variadics using Heterogeneous Lists.  This is one way to
@@ -90,8 +90,8 @@ fn examples() -> LuaResult<()> {
 
     let check_equal = lua.create_function(|lua, args| {
         // Functions wrapped in lua receive their arguments packed together as
-        // LuaMultiValue.  The first thing that most wrapped functions will do
-        // is "unpack" this LuaMultiValue into its parts.  Due to lifetime type
+        // MultiValue.  The first thing that most wrapped functions will do
+        // is "unpack" this MultiValue into its parts.  Due to lifetime type
         // signature limitations, this cannot be done automatically from the
         // function signature, but this will be fixed with ATCs.  Notice the use
         // of the hlist macros again.
@@ -99,7 +99,7 @@ fn examples() -> LuaResult<()> {
 
         // This function just checks whether two string lists are equal, and in
         // an inefficient way.  Results are returned with lua.pack, which takes
-        // any number of values and turns them back into LuaMultiValue.  In this
+        // any number of values and turns them back into MultiValue.  In this
         // way, multiple values can also be returned to Lua.  Again, this cannot
         // be inferred as part of the function signature due to the same
         // lifetime type signature limitations.
@@ -110,7 +110,7 @@ fn examples() -> LuaResult<()> {
     // You can also accept variadic arguments to rust callbacks.
 
     let join = lua.create_function(|lua, args| {
-        let strings = lua.unpack::<LuaVariadic<String>>(args)?.0;
+        let strings = lua.unpack::<Variadic<String>>(args)?.0;
         // (This is quadratic!, it's just an example!)
         lua.pack(strings.iter().fold("".to_owned(), |a, b| a + b))
     });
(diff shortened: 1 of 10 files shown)


Timidger:
In case you didn't know: if you're ok with the possible (but probably not very bad) confusion, you can do `use ::lua::String as LuaString` in the modules where it's used already, and that will be much easier to solve than replacing every instance with ::lua::String.

jonas-schievink:
Good idea!

kyren:
So, I want to talk through this PR, because I don't necessarily disagree with it, but it's definitely not what I would normally do and I'm wondering if maybe I have bad rust style?  So, I obviously use this crate a LOT in a current Chucklefish project, and my pattern so far has always been to do unqualified imports like ```use rlua::{LuaValue, LuaTable, Lua}``` or similar, and then directly use the types.  I really don't like this though, and what I'd much rather be doing is ACTUALLY ```use rlua::*``` even though I know it's "bad".

So with this PR, I guess the crate would be designed to be imported qualified like `use rlua` only, and instead the pattern would be to always reference every type by rlua::Value, rlua::Table, rlua::Lua etc.  For whatever reason, I don't really like that style as much.  I guess find it to be quite a lot of typing and everything is suddenly three word jumps in vim instead of one.

I think, if I accepted this PR the FIRST thing I would do is make a prelude module and re-export everything with a Lua prefix, and then just universally use that.  Maybe that's a GOOD thing though?  What are your opinions?

Also, I'm sorry I made the PR conflict :(

Timidger:
@kyren Well, this is tricky since 1) this is bindings to language bindings, so you're running into easy conflicts like the mentioned `std::string::String` vs `rlua::String` issue and 2) while there's an effort to standardize what's "correct" Rust it's still not as strict as it is in a more established language (e.g Java, C, etc.). Saying that, I think you should definitely not put off considering this change, since now while the library is young and <1.0 is the time to decide on huge breaking changes like this.

I personally have no issue with the prefix, I just find that because of the namespaces that I can usually elide them when making libraries / types. Combined with the strong, static typing and I feel much more comfortable doing that than in e.g Python. 

If you want to do the "rustic" thing, I would suggest looking at [these guidelines](https://aturon.github.io/) (which I _think_ are up-to-date, but you might want to double check if there's a better source) and at the top crates on [crates.io](https://crates.io/). Another good thing would be to see how other language bindings deal with this (e.g [libc](https://crates.io/crates/libc) which seems to take the stance that they'll prefix the types when it's obvious it's an FFI type, which seems to indicate `LuaString` would imply that it's used for FFI when it's really not and is just a Rust-level wrapper. It's your call if you think people will actually be confused by this though).

Ultimately it's your project, and personally I don't care what you choose for this ultimately bike-shed level proposal, but I suggest you reach out to the community (e.g IRC and maybe reddit) to see how they feel about this. I feel like this is the most promising Lua library for Rust (hlua seems all but abandoned, and has limitations that are easy to run into) so it's prudent to make it seem as "rustic" as possible if you want it to achieve that status :)

jonas-schievink:
@Timidger The guidelines you linked look [abandoned](https://github.com/rust-lang/rust-guidelines). The [Rust API guidelines](https://github.com/brson/rust-api-guidelines) are much more recent as they've been recently used to review some popular crates for the libz blitz or something.

@kyren I agree with you that having to use `rlua::Table` everywhere is annoying. I'd probably just import `Table` and use that, since the name is still unambiguous (this kind of datastructure is always called a map in Rust-land). Even for `String` you can always do `use rlua::String as LuaString`, like @Timidger mentioned. Of course, conflicts with `String` are bound to happen all the time. AFAIK those renaming imports are actually idiomatic to do whenever there are colliding names - I've seen `use std::io::Result as IoResult` more than once, even though I prefer `use std::io` and use `io::Result` in that case.

> I think, if I accepted this PR the FIRST thing I would do is make a prelude module and re-export everything with a Lua prefix, and then just universally use that. Maybe that's a GOOD thing though? What are your opinions?

This seems like a not very useful duplication of the API to me. Just doing `use rlua::Thing as LuaThing` when there's a conflict seems like a better idea.

Don't worry about the conflict by the way, but the name `LuaLightUserData` seems *very* awkward to me, so I'm now sure that landing this PR is a good idea :)

kyren:
> This seems like a not very useful duplication of the API to me. Just doing use rlua::Thing as LuaThing when there's a conflict seems like a better idea.

I'm not sure I like that, something doesn't sit well with me when you have names for types like "Integer", or "Number", or "Function", even if they don't conflict.. or even ESPECIALLY if they don't conflict.  Obviously when you're writing code related to Lua it's obvious what you mean but when this is inside something else as part of bindings it just looks really confusing.

I'm going to take a random real use statement from my project:
```
use rlua::{Lua, LuaResult, LuaNil, LuaString, LuaTable, LuaUserData, LuaValue, LuaUserDataType,
           LuaUserDataMethods, FromLua, LuaExternalError, LuaExternalResult};
```
with this change, I think I would at minimum write this:
```
use rlua::(Lua, Result as LuaResult, Nil, String as LuaString, Table as LuaTable, UserData, Value as LuaValue,
           UserDataType, UserDataMethods, FromLua, ExternalError as LuaExternalError, 
           ExternalResult as LuaExternalResult};
```
for consistency.  I don't think this example is especially bad either, because I didn't import Function, or Thread, or Error, or Number, or Integer.

I don't think I mind making the change internally to the crate if it's annoying to some people, but I think I'd rather have a prelude module or some re-export module because I'm just going to write that every time.  I know that things like io::Error are standard, but when so many of the types are conflicting or.. I guess "sound conflicting" or "sound fundamental", it's just really confusing to me.

I'm willing to compromise though, especially if most people would use the crate very sparingly and would rather import unqualified and it doesn't bother them, or would rather import qualified.  It's certainly less typing to say rlua::Something than it is rlua::LuaSomething.

jonas-schievink:
I agree that that's a lot to type out, but I think providing an idiomatic API is more important than that.

A prelude module is probably a good idea anyways (there's a handful of traits we can put in there).

I was also thinking about putting all userdata-related things into a module, since those aren't needed all the time. And if you *do* need them you'll probably need *all* of them so you can just `use rlua::userdata::*;`. Not sure how true that is in practice, though.

kyren:
> I was also thinking about putting all userdata-related things into a module, since those aren't needed all the time. And if you do need them you'll probably need all of them so you can just use rlua::userdata::*;. Not sure how true that is in practice, though.

I actually do want to split lua.rs up, but there's a small annoying bit that I haven't gotten around to dealing with, is that everything in lua.rs has access to the inner Lua state and is unsafe, and everything in other safe modules does not (conversion.rs, multi.rs).  I guess I can use pub restricted to split it up, but it feels like it should be organized better and I just haven't figured out how yet.

I'm aware I could re-export things in a different way than they are internally though.

jonas-schievink:
Turns out there's a conventions RFC about this: https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md

kyren:
I think I've decided that you're correct, or at least most likely correct, and I'd like to merge this and immediately make a prelude module.  Should I do this as a pull request to your branch maybe?

jonas-schievink:
@kyren You can just push to my branch if you want to amend the PR.

Also, I have no clue why Travis is failing. `tests.rs` has no line 816.

jonas-schievink:
Force-pushed a rebased PR which should fix Travis (Travis tests post-merge apparently!).

PR #22, opened 2017-07-23
Add a test ensuring that userdata is dropped

jonas-schievink:
While reviewing 36134e6373bbdfa7ad6c787f1114c74329c902ad I noticed that this wasn't tested anywhere, which is probably bad.


kyren:
This is super important, I was thinking about the fact that this wasn't tested last night but I was too frustrated to do anything about it.  Thank you for adding that!

PR #23, opened 2017-07-24
Beef up the String API

jonas-schievink:
Adds `as_bytes` to view the string as a `[u8]`. Unlike the conversion to a
`&str` slice, this cannot fail.

Adds tests for both functions, which made me notice that `to_str` is broken
when the string contains null bytes, so I made it use the
`as_bytes` method.


kyren:
Yeah, this makes sense!

PR #24, opened 2017-07-25
Enhance documentation and add more examples

jonas-schievink:
There's nothing like using a well-documented crate. Let's get to that point!


kyren:
This is really excellent documentation work!

PR #27, opened 2017-07-25
Few more examples and docs

jonas-schievink:


kyren:
If you have a single type you don't have to use HList, you can just use the type.

jonas-schievink:
Ah, of course. Not sure how I didn't notice that :D

I've split it up in 2 examples

kyren:
Can we change the MyUserData example as well?  I actually sort of dislike the whole HList pattern in general, and I'm currently thinking about how exactly we should get rid of it.  I don't want to give people the idea that they have to use it, at least for simple things.

jonas-schievink:
Ok, done.

Removing hlist would be nice, yeah. But doing that seems to require variadic generics, no?

kyren:
I'll make a proper PR about this eventually, I'm just busy with other parts of work this week so it might be this weekend,  but there's a couple of interrelated changes we could make that would help this situation and *possibly* mean that we could remove HLists from the public API.

The first step would be that we could change the signature of things like this:
```
    pub fn create_function<'lua, F>(&'lua self, func: F) -> Function<'lua>
    where
        F: 'lua + FnMut(&'lua Lua, MultiValue<'lua>) -> Result<MultiValue<'lua>>,
```
to this:
```
    pub fn create_function<'lua, A, R, F>(&'lua self, func: F) -> Function<'lua>
    where
        A: 'lua + FromLua<'lua>,
        R: 'lua + ToLua<'lua>,
        F: 'lua + FnMut(&'a Lua, A) -> Result<R>,
```

I think I've mentioned to you before actually that this is what the API *used* to be, but I had convinced myself (wrongly) that there was a lifetime problem with having the signature of LuaCallback be what it is now.  Since that problem isn't there, we could go back to having this be the API for callbacks, so you could do things like this:
```
let print_person = lua.create_function(|lua, hlist_pat![name, age]: HList![String, u8]| {
    println!("{} is {} years old!", name, age);
});
```
That obviously doesn't get rid of HList, but it helps.  The downside there is that you always have to provide a type for 'args' in callbacks, and you no longer can just type _, you now have to do something like _: HList![], or even something cheeky like _: ().  Still, I think this is probably an improvement?

But, I could go further.  I could resolve myself to not waiting on variadic generics, and simply use tuples everywhere HList exists, and generate the trait impls with macros like other people seem to be doing instead of waiting on variadics.  I could *also* do something slightly more clever, and make.. tuple -> HList instances in the hlist_macro crate, and then I could actually just.. support both tuples and hlists?  Also, we actually internally use hlist_macro for several other things, so this would allow us to just eschew hlists in several interfaces, and they just become a stepping stone from tuples that are slightly easier to deal with, and we only have to write the gross macros once.

Then, the theoretical callback would be:
```
let print_person = lua.create_function(|lua, (name, age): (String, u8)| {
    println!("{} is {} years old!", name, age);
});
```
And then we just loudly explain in the API docs that tuples are how we express LuaMultiValue.  Unpacking a return value as (u8, u8) for example would *only* work with lua doing `return 1, 2`, and never `return {1, 2}`.

But!  Okay, so now we're to a point where we have tuples as argument and return types, we could just use fn traits and automatically deduce all this!  At least I think?  I'm not actually sure of the status on all this and whether it works on stable, but IF we did all this MAYBE we could get to the ideal:
```
let print_person = lua.create_function(|lua, name: String, age: u8| {
    println!("{} is {} years old!", name, age);
});
```

So, after all that I *think* we actually end up pretty close to the same strategy that hlua takes, but I'm not 100% sure.  Also, it appears that hlua still needs functionN macros, and I'm not sure whether that requirement can be removed or not, but maybe it would be better than requiring tuples.

So, I'm getting pretty antsy to make a new version of 'rlua' because there have been a lot of API changes and bugfixes, but I'm sort of waiting on getting time to do all this before considering it, and I want a lot of feedback first.

That's my plan anyway.

kyren:
Also, again thanks for the excellent documentation work!

jonas-schievink:
Thanks for the detailed writeup!

`|lua, name: String, age: u8|` sadly isn't possible on stable Rust today since you must always specify all arguments when using the `Fn*` traits. The rest should be doable, though.

jonas-schievink:
Thinking more about this, we could allow functions to return any type `R` that implements `ToLuaMulti` since we no longer have to thread the conversion results from the `pack`/`unpack` through everything. This makes infallible functions easier to write and makes the API more similar to hlua's.

kyren:
There are two different conflicting ideas here though, fallible functions that should result in a call to lua_error, and infallible functions that return like local res, err = erroring_fn().  Are you proposing that we ONLY have the second kind of function, or something else?

jonas-schievink:
Oh, I was thinking that Lua errors are equivalent to Rust's panics, but maybe that's not the case currently? If they were, these could coexist, I think.

kyren:
Here's the current error handling situation, I'm not sure which parts you're familiar with, so apologies if I'm insulting your intelligence here:

Right now the behavior is that a rust panic is absolutely 100% uncatchable from lua, it will bubble up the different call stacks and the panic be resumed in the main program.  It is only "catchable" with catch_unwind in rust, but the usual caveats of panics not being for error handling apply.

If lua calls the lua error function, then this turns into an Err return when doing the Function::call in rust, this is catchable in both lua (as pcall), and in rust, where you simply handle the Err.

If a rust callback returns Err, this is, symmetrically, the same as a lua function calling error.  Lua can catch this (again with pcall), and also back on the rust side it just turns back into Err.

There is also the Result<T, E> ToLuaMulti impl that you made, which does none of this and simply returns a pair of values to lua.

There is an important note here that through several mechanisms, lua can get at wrapped rust errors, but ONLY the Err kind of error, NEVER the panic kind of error, and this is all sort of complicated to make work right.  This is the reason that there is a special userdata value just for rust errors that implicitly clones, and special versions of pcall and xpcall.

It's pretty confusing and I don't like that it's confusing, but it more or less covers every error handling strategy you would realistically want to do?  It's not inconsistent at least, it's just complex.

jonas-schievink:
Great writeup! I'll put that into some docs if I find the motivation.

What you wrote also makes a lot of sense, since I think there would be little value to being able to catch Rust panics in Lua (kinda seems like a bad idea if untrusted scripts can catch panics and exploit an inconsistent state).

kyren:
To be clear, the only thing that should cause a rust panic with rlua is a deeper rust panic.  It would actually be massively SIMPLER for the API to simply panic on lua 'error', but that is definitely not what I would otherwise want.

I think this is the strategy that hlua was also going for, that script errors turn into Err, but the problem with hlua is that it doesn't take this strategy far enough.  Since lua errors are longjmps, it works in hlua as long as there are no rust callbacks at all, but as soon as there is a rust callback that then calls back *into* lua again, the longjmp would skip over rust frames.  Come to think of it, maybe hlua is sound because it makes this impossible?  I'm not actually sure, because I never used hlua extensively, only really only enough to realize the limitations it has and that it sort of went in a fundamentally different direction.

kyren:
> What you wrote also makes a lot of sense, since I think there would be little value to being able to catch Rust panics in Lua (kinda seems like a bad idea if untrusted scripts can catch panics and exploit an inconsistent state).

Yeah, remember that this library is, I mean it's general purpose sure but it's tuned to Chucklefish's needs, which are complex game engines with a very rich lua api handling lots of interconnecting possibly gross hacky scripts .  Our ideal lua API is a perfect firewall on the scripting layer, giving the scripting layer the ability to *hide* engine bugs would be exactly the opposite of what we'd want haha.  I guess in our use view, the engine is where the "good" code goes and the scripts are where we put most of the hacks :D

I'm not sure we'll ever get to the point where scripts REALLY can't even cause a DoS attack, but that's the goal at least.

kyren:
I thought about it some more, and I can definitely see why you'd think that it would work the way you originally thought.  I'm repeating myself, but I want to talk this out to make sure that it makes sense.

Strategy 1:
```
[Lua 'error']         <->   [rust panic]
[t, e = error_fn()]   <->   [rust Result<T, E>]
```

This is the most obvious strategy,  Lua error is just a longjmp, and rust panics are unwinds with drops, and alternately you can just return either the result or the error in the most idiomatic way for that language.  This WOULD work, you could definitely design an API this way, at the pcall boundary a Lua error would turn into a rust panic, and then on the callback boundary a rust panic would turn back into a Lua error (longjmp).  There are major problems with this approach though, and I made a weak argument above based on game engine design, but there is actually a much much simpler argument based on a simple guiding fact:

_rust panics are not for error handling_

An API like this would have two major effects.  One, the only way to catch something as simple as a script syntax error would be catch_panic.  You COULD make it so that only certain types of lua errors resume as panics on the rust side, but I think this quickly becomes inconsistent and is imo deeply strange.  The second problem is that _Lua becomes capable of catching rust errors_, which like you said feels really wrong.  Again, you could make it impossible for Lua to catch rust panics, but again this starts to not match up to Lua's ideas of error handling again.  The problem is there are two opposing designs, lua errors are designed to be caught and used for error handling, and rust panics are not.

So, this library takes Strategy 2:
```
[invisible]           <->    [rust panic]
[Lua 'error']         <->    [rust Result<T, rlua::Error>]
[t, e = error_fn()]   <->    [rust Result<Result<T, E>, rlua::Error>]
```
In other words, we are sort of lifting error handling one level so that we can remove the ability of lua to even interact with a rust panic.  Now, we have two levels of lua errors to deal with in rust, and currently the encoding is Result<T, rlua::Error> and.. Result<Result<T, E>, rlua::Error>.  This is the complicated, sort of ugly part of Strategy 2.

We COULD simply make another axis of function types so that you either pass in a fallible callback or an infallible one, and that would allow you to eschew Result<Result<T, E1>, E2>, which might help the cognitive burden a bit, or it just may shift it.  We could also add another type in the mix somehow for "lua error pairs" that is not Result, and remove the impl for normal Result.

I'm interested if you have any input on what would be the best strategy here to make this more approachable and understandable.

jonas-schievink:
I definitely agree now that strategy 2 sounds the most reasonable. I currently can't think of a better way to express `t, e = error_fn()` in Rust, so for now the best way to make it understandable seems to be to document it and provide examples showing how which types are translated.

PR #29, opened 2017-07-28
Document `Error`

jonas-schievink:
I didn't yet document *everything* there is to say (in particular, how
exactly custom Rust errors can be passed through Lua), but I've some
changes to this type in mind that I'll do next.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use std::error::Error as StdError;
 use std::result::Result as StdResult;
 
+/// Error type returned by rlua methods.
 #[derive(Debug, Clone)]
 pub enum Error {
     /// Lua syntax error, aka `LUA_ERRSYNTAX` that is NOT an incomplete statement.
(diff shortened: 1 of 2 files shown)


jonas-schievink:
One thing about error handling I still don't understand: What are `ExternalError` and `ExternalResult` used for? `Error::external` doesn't use any of them...

kyren:
> One thing about error handling I still don't understand: What are ExternalError and ExternalResult used for? Error::external doesn't use any of them...

They're extension traits to deal with a specific problem that you run into when you want to use this crate in a large project.  Very often you find yourself wanting to bind things that throw errors unrelated to Lua or rlua, but the error type of callbacks is always rlua::Error.

So, two things were added, the ExternalError variant on rlua::Error, and the two extension traits ExternalError and ExternalResult.  Picking a random example here, but they're used like:
```
lua.pack(is.keys_down.contains(&key.parse().to_lua_err()?))
```
where you just want to, in basically any way you can, shove an error into rlua.

rlua::Error also REALLY needs to be Send for our purposes, but also it becomes really impractical to deal with if wrapped errors are Send but not Sync.  Specifically, since Lua can hold an rlua::Error, and this is more or less unavoidable, you have to use Arc to hold the external error so you can get the error back out of Lua without magically "invalidating" it.  This is the reason for Error + Send + Sync bounds on the external errors.

kyren:
ExternalResult and ExternalError should probably have better names.

kyren:
Also, thank you *again* for your great documentation work.

PR #32, opened 2017-07-31
Major API changes before 0.9

kyren:
First of all, I do NOT intend to merge this pull request as is, it was done very quickly and without updating documentation, this is only for the purposes of discussion.

This PR contains two major API changes, and I want to discuss whether one or both of these should be merged in before 0.9.  First, HLists are no more, and are instead replaced entirely with tuples.  Second, the argument and return types of callbacks are now generic, and the packing / unpacking is done automatically as the callback is called.

I'm not sure whether EITHER of these changes makes for a nicer API, really.  They make it possible *some of the time* to have to type less lines, but I'm not sure either really make things more readable or understandable.

I feel like there's a nicer API hiding in here somewhere though, and I'd like to brainstorm about ways I could change this to be nicer and fix some of the problems.

Things about this API change that I don't like:

Underscore soup and `_: ()`
```lua.create_function(|_, _: ()| <whatever>)```

The fact that everything is now tuples more sharply highlights some of the limitations and silliness 
```let v: f32 = func.call(())?;```
```func.call::<_, ()>((a, b))?;```

Anyway, maybe my concerns are overblown, and this API is a strict improvement?

### DIFF
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -1,5 +1,3 @@
-#[macro_use]
-extern crate hlist_macro;
 extern crate rlua;
 
 use std::f32;
@@ -82,9 +80,7 @@ fn examples() -> Result<()> {
     // This API handles variadics using Heterogeneous Lists.  This is one way to
     // call a function with multiple parameters:
 
-    print.call::<_, ()>(
-        hlist!["hello", "again", "from", "rust"],
-    )?;
+    print.call::<_, ()>(("hello", "again", "from", "rust"))?;
 
     // You can bind rust functions to lua as well
 
@@ -95,7 +91,7 @@ fn examples() -> Result<()> {
         // signature limitations, this cannot be done automatically from the
         // function signature, but this will be fixed with ATCs.  Notice the use
         // of the hlist macros again.
-        let hlist_pat![list1, list2] = lua.unpack::<HList![Vec<String>, Vec<String>]>(args)?;
+        let (list1, list2) = lua.unpack::<(Vec<String>, Vec<String>)>(args)?;
 
         // This function just checks whether two string lists are equal, and in
         // an inefficient way.  Results are returned with lua.pack, which takes
@@ -141,20 +137,20 @@ fn examples() -> Result<()> {
 
     impl UserData for Vec2 {
         fn add_methods(methods: &mut UserDataMethods<Self>) {
-            methods.add_method("magnitude", |lua, vec, _| {
+            methods.add_method("magnitude", |lua, vec, _: ()| {
                 let mag_squared = vec.0 * vec.0 + vec.1 * vec.1;
                 lua.pack(mag_squared.sqrt())
             });
 
             methods.add_meta_function(MetaMethod::Add, |lua, params| {
-                let hlist_pat![vec1, vec2] = lua.unpack::<HList![Vec2, Vec2]>(params)?;
+                let (vec1, vec2) = lua.unpack::<(Vec2, Vec2)>(params)?;
                 lua.pack(Vec2(vec1.0 + vec2.0, vec1.1 + vec2.1))
             });
         }
     }
 
     let vec2_constructor = lua.create_function(|lua, args| {
-        let hlist_pat![x, y] = lua.unpack::<HList![f32, f32]>(args)?;
+        let (x, y) = lua.unpack::<(f32, f32)>(args)?;
         lua.pack(Vec2(x, y))
     });
     globals.set("vec2", vec2_constructor)?;
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -77,29 +77,28 @@ fn examples() -> Result<()> {
     let print: Function = globals.get("print")?;
     print.call::<_, ()>("hello from rust")?;
 
-    // This API handles variadics using Heterogeneous Lists.  This is one way to
-    // call a function with multiple parameters:
+    // This API handles variadics using tuples.  This is one way to call a function with multiple
+    // parameters:
 
     print.call::<_, ()>(("hello", "again", "from", "rust"))?;
 
-    // You can bind rust functions to lua as well
-
+    // You can bind rust functions to lua as well.  Callbacks receive the lua state itself as their
+    // first parameter, and the arguments given to the function as the second parameter.  The type
+    // of the arguments can be anything that is convertible from the set of parameters, in this
+    // case, the function expects two string sequences.
     let check_equal = lua.create_function(|_, (list1, list2): (Vec<String>, Vec<String>)| {
-        // This function just checks whether two string lists are equal, and in
-        // an inefficient way.  Results are returned with lua.pack, which takes
-        // any number of values and turns them back into MultiValue.  In this
-        // way, multiple values can also be returned to Lua.  Again, this cannot
-        // be inferred as part of the function signature due to the same
-        // lifetime type signature limitations.
+        // This function just checks whether two string lists are equal, and in an inefficient way.
+        // Lua callbacks return rlua::Result, an Ok value is a normal return, and an Err return
+        // turns into a Lua 'error'.  Again, any type that is convertible to lua may be returned.
         Ok(list1 == list2)
     });
     globals.set("check_equal", check_equal)?;
 
-    // You can also accept variadic arguments to rust callbacks.
+    // You can also accept runtime variadic arguments to rust callbacks.
 
     let join = lua.create_function(|_, strings: Variadic<String>| {
         // (This is quadratic!, it's just an example!)
-        Ok(strings.0.iter().fold("".to_owned(), |a, b| a + b))
+        Ok(strings.iter().fold("".to_owned(), |a, b| a + b))
     });
     globals.set("join", join)?;
 
@@ -139,9 +138,7 @@ fn examples() -> Result<()> {
         }
     }
 
-    let vec2_constructor = lua.create_function(|_, (x, y): (f32, f32)| {
-        Ok(Vec2(x, y))
-    });
+    let vec2_constructor = lua.create_function(|_, (x, y): (f32, f32)| Ok(Vec2(x, y)));
     globals.set("vec2", vec2_constructor)?;
 
     assert!(
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -127,7 +127,7 @@ fn examples() -> Result<()> {
 
     impl UserData for Vec2 {
         fn add_methods(methods: &mut UserDataMethods<Self>) {
-            methods.add_method("magnitude", |_, vec, _: ()| {
+            methods.add_method("magnitude", |_, vec, ()| {
                 let mag_squared = vec.0 * vec.0 + vec.1 * vec.1;
                 Ok(mag_squared.sqrt())
             });
--- a/examples/examples.rs
+++ b/examples/examples.rs
@@ -84,31 +84,22 @@ fn examples() -> Result<()> {
 
     // You can bind rust functions to lua as well
 
-    let check_equal = lua.create_function(|lua, args| {
-        // Functions wrapped in lua receive their arguments packed together as
-        // MultiValue.  The first thing that most wrapped functions will do
-        // is "unpack" this MultiValue into its parts.  Due to lifetime type
-        // signature limitations, this cannot be done automatically from the
-        // function signature, but this will be fixed with ATCs.  Notice the use
-        // of the hlist macros again.
-        let (list1, list2) = lua.unpack::<(Vec<String>, Vec<String>)>(args)?;
-
+    let check_equal = lua.create_function(|_, (list1, list2): (Vec<String>, Vec<String>)| {
         // This function just checks whether two string lists are equal, and in
         // an inefficient way.  Results are returned with lua.pack, which takes
         // any number of values and turns them back into MultiValue.  In this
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by jonas-schievink** [comment]
I think `_: ()` can always be replaced by just the pattern `()`

File: examples/examples.rs [hunk 1]
```diff
@@ -82,37 +80,26 @@ fn examples() -> Result<()> {
     // This API handles variadics using Heterogeneous Lists.  This is one way to
     // call a function with multiple parameters:
 
-    print.call::<_, ()>(
-        hlist!["hello", "again", "from", "rust"],
-    )?;
+    print.call::<_, ()>(("hello", "again", "from", "rust"))?;
```
jonas-schievink: This is definitely an improvement! The double `((` is a limitation of Rust which I don't think we should work around.

Actually, Rust closures work in much the same way, their arguments are packed into a tuple internally.

examples/examples.rs
```diff
@@ -82,37 +80,26 @@ fn examples() -> Result<()> {
...
-    let check_equal = lua.create_function(|lua, args| {
-        // Functions wrapped in lua receive their arguments packed together as
-        // MultiValue.  The first thing that most wrapped functions will do
-        // is "unpack" this MultiValue into its parts.  Due to lifetime type
-        // signature limitations, this cannot be done automatically from the
-        // function signature, but this will be fixed with ATCs.  Notice the use
-        // of the hlist macros again.
-        let hlist_pat![list1, list2] = lua.unpack::<HList![Vec<String>, Vec<String>]>(args)?;
-
+    let check_equal = lua.create_function(|_, (list1, list2): (Vec<String>, Vec<String>)| {
```
jonas-schievink: Ah, this is where the "underscore soup" now comes from. Couldn't we get rid of the first parameter anyways? If a closure needs it, it can always borrow the `lua` from the outside.

examples/examples.rs
```diff
@@ -82,37 +80,26 @@ fn examples() -> Result<()> {
...
         // This function just checks whether two string lists are equal, and in
         // an inefficient way.  Results are returned with lua.pack, which takes
         // any number of values and turns them back into MultiValue.  In this
         // way, multiple values can also be returned to Lua.  Again, this cannot
         // be inferred as part of the function signature due to the same
         // lifetime type signature limitations.
-        lua.pack(list1 == list2)
+        Ok(list1 == list2)
```
jonas-schievink: Very nice!

examples/examples.rs
```diff
@@ -82,37 +80,26 @@ fn examples() -> Result<()> {
...
     globals.set("check_equal", check_equal)?;
 
     // You can also accept variadic arguments to rust callbacks.
 
-    let join = lua.create_function(|lua, args| {
-        let strings = lua.unpack::<Variadic<String>>(args)?.0;
+    let join = lua.create_function(|_, strings: Variadic<String>| {
         // (This is quadratic!, it's just an example!)
-        lua.pack(strings.iter().fold("".to_owned(), |a, b| a + b))
+        Ok(strings.0.iter().fold("".to_owned(), |a, b| a + b))
```
jonas-schievink: The only thing I think could be improved is the need to do `.0` here. But that's preexisting anyways.

examples/examples.rs
```diff
@@ -141,21 +128,19 @@ fn examples() -> Result<()> {
...
+                Ok(Vec2(vec1.0 + vec2.0, vec1.1 + vec2.1))
             });
         }
     }
 
-    let vec2_constructor = lua.create_function(|lua, args| {
-        let hlist_pat![x, y] = lua.unpack::<HList![f32, f32]>(args)?;
-        lua.pack(Vec2(x, y))
+    let vec2_constructor = lua.create_function(|_, (x, y): (f32, f32)| {
+        Ok(Vec2(x, y))
```
jonas-schievink: Yes, yes, this is definitely the way to go. `unpack`/`pack` and the HList macros add quite a bit of noise, this is very readable.

src/multi.rs
```diff
@@ -83,48 +69,79 @@ impl<'lua, T: FromLua<'lua>> FromLuaMulti<'lua> for Variadic<T> {
...
+impl_tuple! { A B C D E F G }
+impl_tuple! { A B C D E F G H }
+impl_tuple! { A B C D E F G H I }
+impl_tuple! { A B C D E F G H I J }
+impl_tuple! { A B C D E F G H I J K }
+impl_tuple! { A B C D E F G H I J K L }
+impl_tuple! { A B C D E F G H I J K L M }
+impl_tuple! { A B C D E F G H I J K L M N }
+impl_tuple! { A B C D E F G H I J K L M N O }
+impl_tuple! { A B C D E F G H I J K L M N O P }
```
jonas-schievink: I'm wondering if these impact compile times at all. Only going up to 8-tuples would probably still be enough (actually, a function taking 8 unnamed arguments is already far too confusing for my taste).

The stdlib seems to go up to 12, so that's another option for consistency.

kyren:
Thank you a lot for all the great feedback!

examples/examples.rs
kyren: Okay, so after thinking about this, I think not only is this not in general possible, but I think it's actually a soundness bug :(

What happens if you insert a callback that borrows &lua.. that's allowed, because the callback must not outlive the 'lua lifetime right.  Well, insert that into the lua state and make sure no other borrows are there, then move lua to some different address in memory, what happens?

I think maybe the lifetime of all the &self parameters on Lua needs to be different than 'lua somehow, ugh I'm going to have to think about this.

jonas-schievink: Hmm, yeah that's no good. If it's too hard to fix "right" then maybe we should go back to requiring `'static`?

kyren: I don't think there is a way to fix this correctly, or at least not without a big redesign to make sure that the top-level Lua object has a stable address, and even if that's possible I feel like it's really easy to get wrong.

I'm going to move functions back to having a static lifetime.

jonas-schievink: Oh, closures are also able to borrow any other outside variables, so that wouldn't really fix it.

jonas-schievink: Some thoughts and notes (you probably already know this, I'm just writing it down to get a better understanding of the issues at hand):

* In part, this is caused because `Lua` is essentially a heterogeneous collection. `Vec<T>` does not have this problem because its type always expresses the borrow if there is one, because it must contain the lifetime of the borrow.
* It might be possible to express the "shortest borrowed lifetime" as a lifetime parameter of `Lua`, which should solve this.
* My experiments indicate that a closure borrowing the outside `lua` will not compile. In particular, [this](https://play.rust-lang.org/?gist=d6474ac18c718bdf037e0f979c3d7c5e&version=stable) works, but not [this](https://play.rust-lang.org/?gist=3f15d51cf9573e14e446f7805163f4e0&version=stable), and I don't see a way to make the second version work. Therefore, the parameter needs to be kept.
* Compiler errors are extremely bad when the lifetimes are wrong. This does not make it pleasant to fix.

We should also use [compiletest](https://github.com/laumann/compiletest-rs) to make sure this doesn't regress to allowing bad programs to compile (there's another commented-out test somewhere that could be converted, too). Sadly, I haven't found that crate to be very pleasant to use, but it's all we've got for now.

kyren: I might be confused, are you saying that making the closure 'static doesn't fix the borrow problem?  That is not what I've found, I'm going to make an issue about this so we can discuss it.

jonas-schievink: No no, that's perfectly fine. I was referring to "make sure that the top-level Lua object has a stable address", which would not fix the issue, which is basically arbitrary unsafe lifetime transmutation.

kyren: > No no, that's perfectly fine. I was referring to "make sure that the top-level Lua object has a stable address", which would not fix the issue, which is basically arbitrary unsafe lifetime transmutation.

Okay, then I think we're in agreement.  I feel like you know more about these lifetime problems than I do so I'm trying to make sure I understand what you're saying.

> It might be possible to express the "shortest borrowed lifetime" as a lifetime parameter of Lua, which should solve this.

So, you could make Lua itself take a lifetime parameter, and all closures would have to outlive the lifetime parameter?  It would just be PhantomData in Lua?  How do you get the lifetime into the callback, the callbacks can't know which Lua they're ultimately being called from except by getting the main state pointer.  I might be misunderstanding what you're saying though.

> My experiments indicate that a closure borrowing the outside lua will not compile. In particular, this works, but not this, and I don't see a way to make the second version work. Therefore, the parameter needs to be kept.

Which parameter needs to be kept, I'm not following.

kyren: > We should also use compiletest to make sure this doesn't regress to allowing bad programs to compile (there's another commented-out test somewhere that could be converted, too). Sadly, I haven't found that crate to be very pleasant to use, but it's all we've got for now.

Yep, actually I had just written the following before I read what you said, so definitely in agreement:

```
// Need to use compiletest-rs or similar to make sure these don't compile.
/*
#[test]
fn should_not_compile() {
    let lua = Lua::new();
    let globals = lua.globals();

    // Should not allow userdata borrow to outlive lifetime of AnyUserData handle
    struct MyUserData;
    impl UserData for MyUserData {};
    let userdata_ref;
    {
        let touter = globals.get::<_, Table>("touter").unwrap();
        touter.set("userdata", lua.create_userdata(MyUserData)).unwrap();
        let userdata = touter.get::<_, AnyUserData>("userdata").unwrap();
        userdata_ref = userdata.borrow::<MyUserData>();
    }

    // Should not allow self borrow of lua, it can change addresses
    globals.set("boom", lua.create_function(|_, _| {
        lua.pack(lua.eval::<i32>("1 + 1", None)?)
    })).unwrap();
}
*/
```

jonas-schievink: > So, you could make Lua itself take a lifetime parameter, and all closures would have to outlive the lifetime parameter? It would just be PhantomData in Lua? How do you get the lifetime into the callback, the callbacks can't know which Lua they're ultimately being called from except by getting the main state pointer. I might be misunderstanding what you're saying though.

I'm just taking a stab in the dark here. The playground links I posted demonstrate what I mean. The PhantomData isn't necessary, I used that to test things, `'a` is the lifetime outlived by all closures stored in the `Container`. Rust will pick the shortest one automatically (I think). I have no idea how this would look when applied to `Lua`.

> Which parameter needs to be kept, I'm not following.

The first argument (`&Lua`) passed to the closures, since there's no way to safely borrow it from the outside.

kyren: Okay, I think I'm starting to follow here, sorry I'm still pretty noobish at these very complex lifetime problems.

> My experiments indicate that a closure borrowing the outside lua will not compile. In particular, this works, but not this, and I don't see a way to make the second version work. Therefore, the parameter needs to be kept.

Just to be clear, the only reason the first one (https://play.rust-lang.org/?gist=d6474ac18c718bdf037e0f979c3d7c5e&version=stable) works is because it's a self borrow, correct?  The compiler wouldn't actually LET you return that c or move that c at all, so that wouldn't be especially useful right?

jonas-schievink: > Okay, I think I'm starting to follow here, sorry I'm still pretty noobish at these complex lifetime problems.

I don't really have a clear understanding either, or I would have more definitive answers for you ;)

> The compiler wouldn't actually LET you return that c or move that c at all

It does, `c` isn't captured by the closure at all. It just passes a reference to itself as an argument to the closure (like rlua does, too).

kyren: > It does, c isn't captured by the closure at all. It just passes a reference to itself as an argument to the closure (like rlua does, too).

OH... I need to actually slow down and read what you wrote, carefully.

I was reading that incorrectly as borrowing 'c', but you're printing the address of 'container'.  The OTHER version prints the address of c, I didn't see that both sections changed, that would explain part of why I was so confused.  Reading comprehension 2/5.

kyren:
Okay, I think I might be ready to merge this, pending discussion.  I added some tangentially related things, and I sort of muddled the purpose of the PR, I apologize.  I think this is a lot better than 0.8.2, and with bugs and things I'm anxious for a 0.9.

kyren:
Oh, and I realize the README probably could use a lot of updates, I will do that separate from this PR I think.

src/tests.rs
```diff
@@ -223,9 +220,7 @@ fn test_rust_function() {
     ).unwrap();
 
     let lua_function = globals.get::<_, Function>("lua_function").unwrap();
-    let rust_function = lua.create_function(|lua, _| {
-        lua.pack("hello")
-    });
+    let rust_function = lua.create_function(|_, _: ()| Ok("hello"));
```
jonas-schievink: Do you want to replace the `_: ()` with just `()`?

kyren: Oh yeah!  I forgot you mentioned that, that's a great idea.  I guess it's allowed because it's an irrefutable pattern, and that's allowed as lambda arguments?  I have never actually used that before!

jonas-schievink: Yeah, most places in Rust that accept an identifier actually work with irrefutable patterns. This includes `let (x, y) = ...`, closure arguments, but also function arguments so you can do something like `fn move(&mut self, (x, y): (i32, i32))` in case you didn't know

kyren: I actually HAVE used it in let bindings and function arguments lots, I just had never used it as a closure parameter without a type, it was the lack of a type that was getting me somehow.  It makes sense now that I think about it.

PR #34, opened 2017-08-01
[WIP] Enhanced errors

jonas-schievink:
I'm not completely happy with the result, but the error messages ought to be a lot better now (still have to add tests for that). In particular, the conversion error variants are a bit too overloaded for my taste. Maybe I should remove a field from them.

Not sure if all of these changes are uncontroversial :)

Also, I'd like to remove the `ErrorError` variant. It *should* only happen if rlua's own message handler throws an error, which cannot happen. The handler *does* invoke a function that can error, `luaL_tolstring`, but only on wrapped errors and panics, where rlua controls the `__tostring` metamethod and ensures that it cannot panic. Is that correct?

### DIFF
--- a/examples/repl.rs
+++ b/examples/repl.rs
@@ -32,7 +32,7 @@ fn main() {
                     );
                     break;
                 }
-                Err(Error::IncompleteStatement(_)) => {
+                Err(Error::SyntaxError { incomplete_input: true, .. }) => {
                     // continue reading input and append it to `line`
                     write!(stdout, ">> ").unwrap();
                     stdout.flush().unwrap();
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -39,9 +39,12 @@ impl<'lua> FromLua<'lua> for Table<'lua> {
     fn from_lua(value: Value<'lua>, _: &'lua Lua) -> Result<Table<'lua>> {
         match value {
             Value::Table(table) => Ok(table),
-            _ => Err(Error::FromLuaConversionError(
-                "cannot convert lua value to table".to_owned(),
-            )),
+            _ => Err(Error::FromLuaConversionError {
+                from: value.type_name(),
+                to: "table",
+                expected: None,
+                message: None,
+            }),
         }
     }
 }
@@ -56,9 +59,12 @@ impl<'lua> FromLua<'lua> for Function<'lua> {
     fn from_lua(value: Value<'lua>, _: &'lua Lua) -> Result<Function<'lua>> {
         match value {
             Value::Function(table) => Ok(table),
-            _ => Err(Error::FromLuaConversionError(
-                "cannot convert lua value to function".to_owned(),
-            )),
+            _ => Err(Error::FromLuaConversionError {
+                from: value.type_name(),
+                to: "function",
+                expected: None,
+                message: None,
+            }),
         }
     }
 }
@@ -73,9 +79,12 @@ impl<'lua> FromLua<'lua> for Thread<'lua> {
     fn from_lua(value: Value<'lua>, _: &'lua Lua) -> Result<Thread<'lua>> {
         match value {
             Value::Thread(t) => Ok(t),
-            _ => Err(Error::FromLuaConversionError(
-                "cannot convert lua value to thread".to_owned(),
-            )),
+            _ => Err(Error::FromLuaConversionError {
+                from: value.type_name(),
+                to: "thread",
+                expected: None,
+                message: None,
+            }),
         }
     }
 }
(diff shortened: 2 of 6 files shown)


kyren:
> Also, I'd like to remove the ErrorError variant. It should only happen if rlua's own message handler throws an error, which cannot happen. The handler does invoke a function that can error, luaL_tolstring, but only on wrapped errors and panics, where rlua controls the __tostring metamethod and ensures that it cannot panic. Is that correct?

It can be generated in pure Lua right, using xpcall?

kyren:
I like the direction this PR is going though, what parts stand out to you as possibly controversial?  Maybe I missed the really controversial bits in my quick look just now, or maybe we're just thinking along the same lines.

jonas-schievink:
> It can be generated in pure Lua right, using xpcall?

Uhhhh that appears to cause almost endless recursion. It even catches C stack overflows and... continues recursion. Wat. At the very end, xpcall returns `false` and "error in error handling". It stays in Lua, though.

```lua
xpcall(error, function(e) print("got error: ", e) error("rethrowing "..tostring(e)) end, "test error")
```

This is with stock Lua 5.3.4.

jonas-schievink:
> I like the direction this PR is going though, what parts stand out to you as possibly controversial? Maybe I missed the really controversial bits in my quick look just now, or maybe we're just thinking along the same lines.

Like I said, the conversion errors seem too complicated for what they're doing. I also wasn't sure if you'd like the merge of `SyntaxError` and `IncompleteStatement` since it means more code to do the same thing.

kyren:
> Uhhhh that appears to cause almost endless recursion. It even catches C stack overflows and... continues recursion. Wat. At the very end, xpcall returns false and "error in error handling". It stays in Lua, though.

Wait, you're saying this happens *outside* of rlua, just in the stock Lua REPL?

jonas-schievink:
Yes! It's amazing 🎉

kyren:
> The handler does invoke a function that can error, luaL_tolstring, but only on wrapped errors and panics, where rlua controls the __tostring metamethod and ensures that it cannot panic. Is that correct?

Just to be clear, I DO believe this is correct.  In any case, any error in a rust error handler would be a bug.

I'm still a bit confused though, when exactly does stock lua generate LUA_ERRERR?  Is it only generated if you set a panic handler with lua_atpanic?

jonas-schievink:
I believe that happens when the message handler set with `lua_pcall` raises an error.

kyren:
> I believe that happens when the message handler set with lua_pcall raises an error.

But wouldn't that also apply to lua xpcall then?  That's what's confusing me, it feels like the current behavior with xpcall is a bug or something.  `safe_xpcall` just does a `lua_call` on the lua provided message handler, which of course can raise an error, so it seems like this could still be generated from lua?  I feel like there's something I'm missing.

jonas-schievink:
Oh, yes, that's right. Is the Lua documentation wrong, then? The code I posted above never causes `LUA_ERRERR` (even when run in rlua's repl), errors in the handler are just caught by the handler *again*. Or maybe it *is* thrown but one of the xpcalls catches it?

kyren:
I think I'd like to merge this PR before 0.9 if we can figure out the situation with ERRERR

kyren:
Also, if we get rid of ERRERR it will cause me to delete my favorite line of code in rlua,

image

jonas-schievink:
It looks like the error handling Lua does effectively convolutes ERRERR and ERRRUN. ERRERR is only raised in very specific situations which makes it not worth to treat differently than other runtime errors in my opinion. Besides that, the Lua manual is clearly wrong, which is going to confuse users further. I'm going to merge the 2 errors instead of aborting on ERRERR.

kyren:
> It looks like the error handling Lua does effectively convolutes ERRERR and ERRRUN. ERRERR is only raised in very specific situations which makes it not worth to treat differently than other runtime errors in my opinion. Besides that, the Lua manual is clearly wrong, which is going to confuse users further. I'm going to merge the 2 errors instead of aborting on ERRERR.

Huh!  I wonder if that's worth a bug report to the Lua people at PUC Rio.

jonas-schievink:
I've already found this thread on the mailing list: http://lua-users.org/lists/lua-l/2010-10/msg00200.html (it's not very enlightening)

kyren:
Maybe it calls the error handler as many times as it is able, and then after all that causes a generic error, and their logic is that that is the best that they can do?  It seems sketchy, but I could see thinking that's the best behavior (give the error handler as many opportunities to not error as we can), but if so they should document that!

PR #35, opened 2017-08-02
Use rustyline for the REPL

jonas-schievink:
This makes the REPL more usable as you can now edit lines and recall
previously executed statements.


kyren:
Neat, I didn't know about rustyline.  I wouldn't want a dependency on the crate to rustyline, but this is okay because it's a dev dependency only.

PR #42, opened 2017-09-09
Switched gcc::Config to gcc:Build in build.rs.

thelearnerofcode:
gcc::Config was replaced by gcc::Build in gcc-rs.


kyren:
Thank you!

PR #45, opened 2017-09-15
Extend String API

jonas-schievink:
The first commit only moves stuff around, as `lua.rs` is getting too big for my taste (>2000 lines). The rest of them contain the functional changes.

Re: Moving string tests to `string.rs` - I think it makes sense to have all functionality acting directly on strings in this file, so maybe I should move the `FromLua` and `ToLua` impls there, too. What do you think?

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -4,6 +4,8 @@ use std::string::String as StdString;
 
 use error::*;
 use lua::*;
+use string::String;
+use table::Table;
 
 impl<'lua> ToLua<'lua> for Value<'lua> {
     fn to_lua(self, _: &'lua Lua) -> Result<Value<'lua>> {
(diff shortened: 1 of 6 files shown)


kyren:
Sorry for taking such a long time to get around to looking at this.  I definitely want to split up lua.rs, in fact the only reason it was so monolithic is because for a while I was avoiding pub(crate).  Clearly having pub(crate) and splitting the modules is the better approach here.

The string additions look really sensible, and I definitely want to include those, good work!

So far, my personal rust style has been to separate out tests into their own files, but seeing how much (quality!) documentation there is for all of the user facing types, it seems weird to include the documentation, implementation, and *not* the tests in the same file.  I'm fine with moving forward with that organization.  I think I also agree with moving the From/To impls there as well.

I think I'm going to merge this, and then finish this organizational change for the rest of the contents of lua.rs as well.

PR #46, opened 2017-09-16
Use compiletest to make sure unsafe things do not compile

jonas-schievink:
I only added 2 tests I found in `tests.rs`, it would be nice to gather more. If anyone finds something unsafe that we should test against, please leave a comment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: rust
-rust:
-  - stable
-  - beta
-  - nightly
 env:
   - RUSTFLAGS="-D warnings"
 matrix:
+  include:
+    - rust: stable
+      script: cargo test --lib -v
+    - rust: beta
+      script: cargo test --lib -v
+    - rust: nightly
   allow_failures:
     - rust: nightly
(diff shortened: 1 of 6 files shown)


jonas-schievink:
Oh yeah, Travis fails since this only works on nightly. Give me a second.

jonas-schievink:
UGH. Why isn't there a more elegant and usable solution for compile-fail tests? They're pretty important for all sorts of crates...

kyren:
I ended up working off of what you had and managed to get travis working as of 1426bdbc164f46af3f68c523a4356177377bac6f, I didn't put compiletest_rs in [dev-dependencies] though, but it is optional?

I'm not sure if I have it set up in the most optimal way.

jonas-schievink:
@kyren Very nice! The only issue with making compiletest a regular dependency is that downstream crates can now request the `compiletest_rs` feature to be enabled, too. I don't think that's too big of a deal, though.

Eventually I'll revisit my plans to make compile-fail tests more accessible, probably after https://github.com/rust-lang/cargo/pull/4734 lands or https://github.com/rust-lang/rfcs/pull/2318 is accepted and implemented.

Also closing, since this is implemented now.

PR #47, opened 2017-09-16
Don't load the base library into the "base" global

jonas-schievink:
The stock Lua interpreter doesn't do this either. AFAIK the "base" library is supposed to be loaded into "_G" only as it contains functions like `assert` and `error`.


kyren:
👍

kyren:
I believe that was actually a copy paste fail on my part, thank you for catching that.

PR #54, opened 2017-10-29
Added ability to load debug lib, Fixes #52

### LINKED ISSUE #52: debug library not included
Hey there,

The awesome lua library (which I'm trying to run with these bindings) makes great use of the `debug` functionality of Lua (specifically `debug.traceback`), but it's not included in these bindings. I could make it myself, but I just wanted to double check to see why it's not included.

Timidger:
Adds unsafe method `load_debug` to `Lua` that allows loading of the debug library. Marked it unsafe and added scary text to dissuade users from using it.

### DIFF
--- a/src/lua.rs
+++ b/src/lua.rs
@@ -563,6 +563,7 @@ impl Lua {
     /// the guarantees of rlua.
     pub unsafe fn load_debug(&self) {
         ffi::luaL_requiref(self.state, cstr!("debug"), ffi::luaopen_debug, 1);
+        ffi::lua_pop(self.state, 1);
     }
 
     /// Loads a chunk of Lua code and returns it as a function.
--- a/src/lua.rs
+++ b/src/lua.rs
@@ -557,6 +557,14 @@ impl Lua {
         }
     }
 
+    /// Loads the Lua debug library.
+    ///
+    /// The debug library is very unsound, loading it and using it breaks all
+    /// the guarantees of rlua.
+    pub unsafe fn load_debug(&self) {
+        ffi::luaL_requiref(self.state, cstr!("debug"), ffi::luaopen_debug, 1);
+    }
+
     /// Loads a chunk of Lua code and returns it as a function.
     ///
     /// The source can be named by setting the `name` parameter. This is generally recommended as it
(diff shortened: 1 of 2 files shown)


src/lua.rs
```diff
@@ -557,6 +557,14 @@ impl Lua {
         }
     }
 
+    /// Loads the Lua debug library.
+    ///
+    /// The debug library is very unsound, loading it and using it breaks all
+    /// the guarantees of rlua.
+    pub unsafe fn load_debug(&self) {
+        ffi::luaL_requiref(self.state, cstr!("debug"), ffi::luaopen_debug, 1);
```
jonas-schievink: You need to add a `ffi::lua_pop(state, 1);` after this, and (I think) check that the stack still has room for the pushed module.

Timidger:
Added a test

src/lua.rs
Timidger: How would I check that hte stack still has room for the pushed module?

Timidger: Fixed pop issue

jonas-schievink: I think the correct way to do that in this case is to call `util::check_stack`

src/tests.rs
```diff
@@ -15,6 +15,23 @@ fn test_load() {
 }
 
 #[test]
+fn test_load_debug() {
+    let lua = Lua::new();
+    lua.eval::<()>("debug", None).unwrap();
```
jonas-schievink: This doesn't do anything regardless of the value of `debug`. Try to just `exec("assert(debug == nil)").unwrap()`

Timidger: Nice catch, fixed

src/lua.rs
Timidger: Added the check before the module loading

Timidger:
What's the status on this? Any more tests I should add / changes I should make to the debug loading code?

kyren:
No, it's fine, I should have already merged it but I was delaying trying to figure out an API for selectively loading all of the other libraries as well.  I think actually that that can wait for 0.10.

I'm gonna go ahead and merge this then, thank you for all the work on this!.

PR #55, opened 2017-11-08
Fixes error when setting metatable to nil

Timidger:
If you set the metatable to nil, currently it tries to get the `__gc` method without first checking whether or not the metatable is defined.

This fixes the issues I raised in #52


kyren:
Ugh, thank you for tracking that bug down.  I’m sorry I never had a chance to look deeper into this, I’m glad you were able to find it.  This fix seems right to me!

PR #61, opened 2017-12-27
Add type conversions for {Hash,BTree}Sets

mmirate:
The Lua convention for a set (*S*&ni;*x&sup1;* ^ *S*&ni;*x&sup2;* ^ ...) is that *S* is a table where each *x* is a key whose value is `true`.

The FromLua and ToLua implementations I'm proposing here, then. abstract-away the `.iter().cloned().map(|k| (k, true))` dance needed to expose to Lua one of the Set types from std::collections.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap};
+use std::collections::{BTreeMap, HashMap, BTreeSet, HashSet};
 use std::hash::{BuildHasher, Hash};
 use std::string::String as StdString;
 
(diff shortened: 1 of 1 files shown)


jonas-schievink:
IIRC those were removed intentionally after the discussion in https://github.com/chucklefish/rlua/issues/11

mmirate:
Ah, interesting. Looks like the old behavior effectively constructed/deconstructed sets from vectors, and the them being removed was due to the perception that the behavior in this PR was merely another valid way to do it.

I disagree, in that this behavior seems far more obvious.

... Though the conversion between a Set and a Map actually ends up doing *more* work than reconstructing a set-ish table from a Vec of the members (because iterating through a set or map is much less trivial than through a Vec). Quite annoying that there isn't a lower-cost way in std::collections to transform a Set + value-function to a Map nor a Map's keys to a Set; but without it, the abstraction I'm proposing does no good. (Nor does, really, using a std::collections Set with rlua in the first place.) *sigh*. Closing.

Issue #60, opened 2017-12-16
Consider hiding `Error` internals


jonas-schievink:
Currently, adding a variant to `Error` (or changing an existing one) is a breaking change. This happens pretty frequently during development. We could prevent that by making `Error` a struct with a few helper methods to create a specific error, and make the internal enum private so it's not part of the API.

This is probably not that important right now, but when rlua gets closer to a 1.0 release it could allow more freedom in expanding rlua's internals (better error messages, more distinguished error cases without a breaking change). I'm just opening this issue to get a bit of feedback.

Timidger:
If you do this, it would make it harder to remember to `match` on specific error cases, especially when there's a change and e.g an error is made more specific.

Though if you're thinking of changing the error system, might I suggest the new [failure crate](https://github.com/withoutboats/failure)?

jonas-schievink:
> If you do this, it would make it harder to remember to match on specific error cases, especially when there's a change and e.g an error is made more specific.

Do you have an example where this is useful? I've mostly returned elaborate error messages from within callbacks (ie. where the error actually happens), not after catching an error that happened elsewhere.

> Though if you're thinking of changing the error system, might I suggest the new failure crate?

That's probably a good idea, but it does pull in an additional dependency.

torkleyy:
The way I avoid this problem usually is to add another variant, `__NonExhaustive` with `#[doc(hidden)]` and a doc comment saying that you shall never match against it, so you need to have a `_` pattern. This would make adding errors a non-breaking change. Changing an error obviously still is a breaking change.

jugglerchris:
Closing as this is out of date (`Error` has changed to use `failure` and then back again in that time).

PR #63, opened 2018-01-24
Experimentally use the `failure` crate for errors

kyren:
I understand this may be a controversial change, which is why I want to discuss this as a PR.

The situation that is driving this change is that internally, Chucklefish projects have migrated away from using the `error_chain` crate to the `failure` crate.  The easiest thing for me to do when only considering Chucklefish's needs would be to merge this PR as is, but the reasons behind it and the total available options are somewhat nuanced.

`rlua` is in a unique, more complicated situation than many other library crates because it must be at once both a consumer of error types from external code, and a producer of internal error types when interpreting Lua, and both of these errors must be possible to represent at the same time.  Also, errors should at minimum be Send, but since for technical reasons around giving the ability for Lua itself to handle errors, they also need to be Clone.  In order for `rlua::Error` to be both Send and Clone, external error types must be in an Arc, and so must ALSO be Sync.

So, the summary is that `rlua::Error` is `Send + Sync + Clone + 'static`, and wrapped external errors must be `Send + Sync + 'static`.  This actually makes the current situation when using the `error_chain` crate problematic, because I believe only version 0.8 of `error_chain` produces Sync errors.  Chucklefish has a patched version of `error_chain` just for this problem, but again we're transitioning away from `error_chain` entirely anyway.

When using the `failure` crate with `rlua`, we ran into a different set of issues.  Inter-operating with `std::error::Error` style errors with `failure` is actually fairly well supported, with an included `Compat` wrapper to produce a `std::error::Error` compatible error from a `failure` type error.  The one caveat here is that the `Compat` wrapper type does not and cannot return a cause in `Error::cause`, which means that a long chain of errors will always stop at the first `failure::Fail` -> `std::error::Error` boundary.  Similarly, all types that are `std::error::Error + Send + Sync + 'static` implement `failure::Fail`, but again since it is impossible for them to return a `failure::Fail::cause`, a  chain of errors will also stop on a transition from `std::error::Error` to `failure::Fail`.  Code using `rlua` is quite likely to end up with "error sandwiches" where error types go from being external (in callbacks) to `rlua::Error`, and eventually back to being some external error type in the application using `rlua`, so these annoying limitations on the transition layers are particularly problematic.

For Chucklefish, using the `failure` crate is quite attractive; the pattern of using custom `failure::Fail` types for small errors and quickly transitioning to `failure::Error` for large application level errors simplifies error handling immensely without reducing any error handling power that we would realistically use.  With this change, either small local errors which implement `failure::Fail`, or large integrated errors `failure::Error` both can be wrapped as an external `rlua::Error`, and since `rlua::Error` also implements `failure::Fail`, there are no breaks in the cause chain.  Additionally, `failure::Error` is a huge performance improvement over the giant trees of enums that `error_chain` tends to produce.

For other people using `rlua` that are not necessarily using `failure`, there are two downsides.  One is that `rlua::Error` no longer can implement `std::error::Error` due to a conflicting impl of `failure::Fail`.  The other is that though `std::error::Error`s can be wrapped as external errors if they are also `Send + Sync + 'static` (same as now), doing so breaks the cause chain.

Are these downsides enough that this is an unacceptable change?  Is there a change that can make this inter-operate with both people using `std::error::Error` AND `failure`?  Should this be behind a feature flag?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ builtin-lua = ["gcc"]
 
 [dependencies]
 libc = { version = "0.2" }
+failure = { version = "0.1.1" }
 
 [build-dependencies]
 gcc = { version = "0.3.52", optional = true }
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,7 +1,8 @@
 use std::fmt;
 use std::sync::Arc;
-use std::error::Error as StdError;
 use std::result::Result as StdResult;
+
+use failure;
 
 /// Error type returned by `rlua` methods.
 #[derive(Debug, Clone)]
@@ -107,7 +108,7 @@ pub enum Error {
     /// Returning `Err(ExternalError(...))` from a Rust callback will raise the error as a Lua
     /// error. The Rust code that originally invoked the Lua code then receives a `CallbackError`,
     /// from which the original error (and a stack traceback) can be recovered.
-    ExternalError(Arc<StdError + Send + Sync>),
+    ExternalError(Arc<failure::Error>),
 }
 
 /// A specialized `Result` type used by `rlua`'s API.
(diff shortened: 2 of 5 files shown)


Timidger:
For Way Cooler, I actually don't use either crate (`error_chain` was too complicated for me, and `failure` came after I put Way Cooler on mostly hibernation until wlroots-rs is complete).

My plan is to move to `failure` since it's planned to be integrated into the std. I have no problem with rlua using it. In fact it's a big push to get me to hurry up and use it in wlroots-rs, so thanks.

jonas-schievink:
If I understood correctly, the `failure` crate is supposed to be the future error handling story for all of Rust (eventually moving into libstd), so I'm definitely in favor of supporting `failure` in as many crates as possible.

kyren:
I was actually unaware that the Fail trait was *definitely* moving into std, the last I heard about it it was still being discussed.  Do you guys have a link to any ongoing discussions?

In any case, you guys seem very positive on this change.  That, combined with the news that the Fail trait is going to become part of libstd is actually a huge relief, because it makes this change pretty non-controversial.

Thank you!

jonas-schievink:
Ah, sorry, I only saw *someone* mention *somewhere* that this is *eventually* the long-term plan.

Since it would be the first addition of this size to libstd, I wouldn't hold my breath for it happening anytime soon even if the plan was more concrete (just because it takes time and is going to be fairly low priority, given that the external crate works just fine).

kyren:
> Ah, sorry, I only saw someone mention somewhere that this is eventually the long-term plan.

Oh, okay.  I still think you've probably heard more about it than me, I think the only discussions I read were on the original `failure` announcement on /r/rust.

In any case, it does seem like the whole ecosystem is moving towards `failure`, and even in the unlikely event that the error handling story goes through another radical change very soon, this crate is still < 1.0.  Also, in the absolute worst case of someone vehemently against using `failure`, the most that this will cause is an inconvenience.

Since this change seems much more non-controversial than I had originally thought, so I'm going to merge it.

kyren:
I completely missed that you guys already discussed this in #60, sorry about that!

Timidger:
No worries, thanks for asking around before making what could have been a controversial change 😄

PR #66, opened 2018-01-26
Clarify a few docs

jonas-schievink:


kyren:
Technically, we COULD allow rust to override __gc now, but it's kind of pointless as you can just use Drop..

I don't think we should mention the safety issues in the docs, maybe we should add a __gc MetaMethod entry just so we don't have to explain why it's not there haha.

kyren:
This is all really good, thank you yet again!

Issue #40, opened 2017-09-01
Why is Lua not Send?


mkpankov:
I believe wrapping `Lua` in `Arc<Mutex<...>>` should allow passing it to other threads and synchronized access?

Right now it's this:

```
error[E0277]: the trait bound `*mut rlua::ffi::lua_State: std::marker::Send` is not satisfied in `rlua::Lua`
  --> src/main.rs:29:29
   |
29 |     let event_loop_thread = thread::spawn({
   |                             ^^^^^^^^^^^^^ `*mut rlua::ffi::lua_State` cannot be sent between threads safely
   |
   = help: within `rlua::Lua`, the trait `std::marker::Send` is not implemented for `*mut rlua::ffi::lua_State`
   = note: required because it appears within the type `rlua::Lua`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Mutex<rlua::Lua>`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<std::sync::Mutex<rlua::Lua>>`
   = note: required because it appears within the type `[closure@src/main.rs:32:9: 55:10 lua:std::sync::Arc<std::sync::Mutex<rlua::Lua>>, barrier:std::sync::Arc<std::sync::Barrier>]`
   = note: required by `std::thread::spawn`
```

jonas-schievink:
Yeah, this should be safe to do. Just needs an impl.

kyren:
Unfortunately, if you do that, you have to universally decide that all userdata inside Lua should also be Send, which is really bad if you want to put an Rc into Lua.  Since there's no way to abstract over whether Lua should be all Send or all !Send, I picked !Send intentionally.  Perhaps there is some way of abstracting this, or maybe macroing this, so you can choose whether you want the Send version of Lua or the !Send version?

jonas-schievink:
Oh, right. That's a bit unfortunate. I guess you could try to create the `Lua` inside the thread if you pass everything needed for construction of your environment, but perhaps that's not always possible?

Anyways, I think we should look at a few use cases where `Lua: Send` makes an important difference.

kyren:
I actually have a solution to this inside the project I'm currently working on that creates a thread JUST for Lua in certain cases where otherwise it would need to be Send, it's extremely unfortunate.  In fact, I have an incredible hack of a wrapper just for it:

https://gist.github.com/kyren/d010f25cc6d98bcfc4044c44518b5676

[...]

Issue #56, opened 2017-12-04
Why does create_function specify 'static?


magnonellie:
When playing around trying to make a prototype for a system I'm developing, I found that I bumped into a wall when using `Lua.create_function`. It has a lifetime bound specified as `'static` which caused trouble, but when I tried changing it to `'lua` it worked and it doesn't seem to have caused trouble elsewhere. So my question is why the boundary is `'static` instead of `'lua`. I'm pretty new to Rust so I may have missed something very obvious here, but I thought I'd ask anyway.

Thanks!

kyren:
We’ve been through this before:

https://github.com/chucklefish/rlua/issues/33
https://github.com/chucklefish/rlua/issues/20

Anything that is passed to lua morally gets put into an Any, and it’s very very difficult to safely

magnonellie:
Yeah, now I feel pretty bad. Those issues completely went over my radar. Sorry for bringing it up again, but thank you for the explanation anyway! Maybe this should be added to some kind of FAQ or something?

jonas-schievink:
We could just add a note to the `create_function` docs

jonas-schievink:
FWIW, we could add a lifetime parameter to `Lua` that specifies how long the longest-lived externally borrowed data lives to make this sound. Given that this has come up a few times already, maybe we should try that? It wouldn't magically solve anything, but maybe it solves a few use cases.

(and this time, I'm very certain that it'll be safe, unlike my last attempt)

kyren:
> Yeah, now I feel pretty bad. Those issues completely went over my radar. Sorry for bringing it up again, but thank you for the explanation anyway! Maybe this should be added to some kind of FAQ or something?

Don't feel bad!  It's actually a really good suggestion, which is why we already tried it :D  These issues are not exactly obvious, either.

kyren:
> FWIW, we could add a lifetime parameter to Lua that specifies how long the longest-lived externally borrowed data lives to make this sound. Given that this has come up a few times already, maybe we should try that? It wouldn't magically solve anything, but maybe it solves a few use cases.

> (and this time, I'm very certain that it'll be safe, unlike my last attempt)

That's really interesting, can you flesh out this plan a bit?

jonas-schievink:
@kyren Sure, I'll try to describe it.

[...]

PR #68, opened 2018-02-07
Lots of changes, not sure if actually safe yet.

kyren:
* Make Lua Send
* Add Send bounds to (nearly) all instances where userdata and functions are
  passed to Lua
* Add a "scope" method which takes a callback that accepts a `Scope`, and give
  `Scope` the ability to create functions and userdata that are !Send, and also
  functions that are not 'static!.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -112,7 +112,7 @@ impl<'lua> FromLua<'lua> for AnyUserData<'lua> {
     }
 }
 
-impl<'lua, T: UserData> ToLua<'lua> for T {
+impl<'lua, T: Send + UserData> ToLua<'lua> for T {
     fn to_lua(self, lua: &'lua Lua) -> Result<Value<'lua>> {
         Ok(Value::UserData(lua.create_userdata(self)?))
     }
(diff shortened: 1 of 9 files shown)


kyren:
This is a big change, and I really feel like I might be missing something re: safety, so it'd be great if anybody wants to take a look at it and tell me how I've inevitably missed some crucial safety aspect.

If this is a viable strategy, though, it fixes #40 and.. more or less #56 as well.

kyren:
A key realization for me was that everywhere that I've ever NEEDED to give a !Send type to Lua, it has always been some sort of lock, like RwLockReadGuard etc.  These also always need some sort of mechanism to manually expire them and ensure that they don't leak out, so in this way, `Lua::scope` does double duty by making it safe for Lua to be Send, as well as making the expiration mechanism built in.  It doesn't seem like a feature, but it is.

kyren:
One thing that's not tested here that should be tested is accessing expired values through `rlua::Function` and `rlua::AnyUserData`.  I believe the behavior will be that if `Function` is called, it will just always error, and `AnyUserData` will behave as though it is not any type at all.

kyren:
I think I'm going to go ahead and merge this into master, but I'd still like to discuss it before doing another major rlua release.

It is potentially the most annoying API change yet because of the new requirement of userdata to be Send, I don't really have a good sense of the amount of pain this will cause.

kyren:
Well, I merged this, but I actually screwed up the lifetimes in `Lua::scope`, and it's possible to reference things that don't live long enough.

I'm currently staring that the crossbeam crate trying to figure out the correct technique.

Issue #67, opened 2018-02-01
Serde Serializer/Deserializer for Lua Value


zrkn:
Currently conversion of Rust types to Lua `Value`s is uses `FromLua` and `ToLua` traits. However, it might be cumbersome to implement these traits manually. One solution is to implement derive procedural macro. However, it is quite a lot of work and introduces extra dependency for end user.

On the other hand, a very similar data type is used in `serde_json` - [Value](https://docs.rs/serde_json/1.0.9/serde_json/enum.Value.html). With `Serializer`/`Deserializer` traits implemented, basically any data type with `Serialize`/`Deserialize` trait can be converted to/from it.

I have test implementation of `Serializer` in [zrkn/rlua](https://github.com/zrkn/rlua/tree/serde) (mostly copied from `serde_json`). In my project I only need to convert Rust values to Lua, so it works for me. However if you would consider merging this feature upstream I can implement `Deserializer` too and make pull request.

kyren:
I've definitely had the same idea, it's cool that you implemented this and I'll definitely take a look at it.

The problem with doing this though, is that it creates two paths for serialization, the Serialize / Deserialize path and the ToLua / FromLua path.

If there a way that we could provide some sort of `to_lua_serialize!` and `from_lua_deserialize` macros that make a struct that is Serialize / Deserialize implement ToLua / FromLua, that would be really interesting.  I haven't looked into it, but I don't think that's possible to do in a way where it forwards to ToLua / FromLua on the fields of the struct implementing Serialize / Deserialize, and if you have two separate serialization systems, I don't think that's going to work out very well.

The problem is something like this, suppose you have a `struct Complicated` with a bespoke `impl ToLua for Complicated` and `impl FromLua for Complicated`.  Then, if you have a `#[derive(Serialize, Deserialize)] struct Compound { complicated: Complicated }`, at best using Serialize / Deserialize to do the Lua conversion won't compile, but at worst it may actually use a serialization that you don't actually want.  I don't really want there to be multiple parallel ways to do Lua serialization.

zrkn:
There is a difference between `ToLua`/`FromLua` conversion path and `Serialize`/`Deserialize`, which can be a reason to keep them separate. `ToLua`/`FromLua` consumes `self`, when `Serialize`/`Deserialize` takes it by reference. Basically, first is _conversion_ and second is _serialization_ which is a bit different operations. End user can choose which is more appropriate for them.

zrkn:
[...]

PR #75, opened 2018-03-09
[Very Experimental] #derive[LuaTable]

kngwyu:
This PR proposes `#[derive(LuaTable)]` syntax.
This attribute implements new `FromTable` and `IntoTable` trait.
We can convert struct with `#[derive(LuaTable)` into `Table` by `into_table` method, and vice versa(by `from_table` method.
See example for usage.

Though there are many problems (e.g. integration with `UserData` type), I think this is useful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ builtin-lua = ["cc"]
 libc = { version = "0.2" }
 failure = { git = "https://github.com/rust-lang-nursery/failure", version = "1.0" }
 compiletest_rs = { version = "0.3", optional = true }
+rlua-derive = { version = "0.0.1", path = "./rlua-derive", optional = true }
 
 [build-dependencies]
 cc = { version = "1.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,6 @@ criterion = "0.2.0"
 [[bench]]
 name = "benchmark"
 harness = false
+
+[workspace]
+members = ["rlua-derive"]
--- a/rlua-derive/Cargo.toml
+++ b/rlua-derive/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "rlua-derive"
+version = "0.0.1"
+authors = ["kyren <catherine@chucklefish.org>", "kngwyu <yuji.kngw.80s.revive@gmail.com>"]
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = "0.12"
+quote = "0.4"
+synstructure = {git = "https://github.com/nrxus/synstructure.git"}
--- a/rlua-derive/Cargo.toml
+++ b/rlua-derive/Cargo.toml
@@ -10,3 +10,6 @@ proc-macro = true
 syn = "0.12"
 quote = "0.4"
 synstructure = {git = "https://github.com/nrxus/synstructure.git"}
+
+[dev-dependencies]
+rlua = {path = "../", features = ["rlua-derive"]}
--- a/rlua-derive/Cargo.toml
+++ b/rlua-derive/Cargo.toml
@@ -9,7 +9,7 @@ proc-macro = true
 [dependencies]
 syn = "0.12"
 quote = "0.4"
-synstructure = {git = "https://github.com/nrxus/synstructure.git"}
+synstructure = {git = "https://github.com/mystor/synstructure.git"}
 
 [dev-dependencies]
 rlua = {path = "../", features = ["rlua-derive"]}
--- a/rlua-derive/examples/derive_usage.rs
+++ b/rlua-derive/examples/derive_usage.rs
@@ -1,0 +1,62 @@
+extern crate rlua;
+
+use rlua::*;
+
+fn derive_usage() -> Result<()> {
+    let lua = Lua::new();
+    #[derive(LuaTable, Debug)]
+    struct Vec2D(f64, f64);
+
+    let table = Vec2D(3.0, 4.0).into_table(&lua)?;
+    lua.globals().set("v2", table)?;
+    lua.exec::<()>(
+        r#"
+print(v2._0)
+print(v2._1)
+v2._0 = 100
+v2._1 = v2._0 * 2
+"#,
+        None,
+    )?;
+
+    let table = lua.globals().get("v2")?;
+    let v2 = Vec2D::from_table(table, &lua)?;
+    println!("{:?}", v2);
+
+    #[derive(LuaTable, Debug)]
+    struct Vec3D {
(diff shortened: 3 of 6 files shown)


jonas-schievink:
This seems similar to https://github.com/chucklefish/rlua/issues/67. Is there a reason why it can't use serde's existing infrastructure (aka `#[derive(Serialize)]`)?

kngwyu:
@jonas-schievink 
As discussed in #67,  `#[derive(Serialize)]` can't integrate with `ToLua/FromLua`, so I thought it's better to have specific implementation for integration with `UserData`.

Though now only `FromTable` and `IntoTable` traits are implemented by `#[derive(LuaTable)]`, we can implement `ToLua` and `FromLua`, or any other traits.
In the future, I think it's good if we can use syntax like 
```rust
// LuaTable defines conversion between Rust struct <=> Lua Table
// LuaProtoType defines constructor in Lua, in this case Vec2D:new() (it's only in my dream!)
#[derive(LuaTable, LuaProtoType, Debug, Default)]
struct Vec2D(f64, f64);
// LuaMethods defines another methods of Vec2D in Lua  (it's only in my dream, too)
impl LuaMethods<'lua> for Vec2D {
    fn add_methods(methods: &mut UserDataMethods<Self>) {
        methods.add_method("length", |_, this, _: ()| {
            Ok((this.0 * this.0 + this.1 + this.1).sqrt())
        });
    }
}
let lua = Lua::new();
lua.globals().set_class::<Vec2D>::();
lua.exec::<()>(
    r#"
v = Vec2D:new(3, 4)
l = v.length()
"#,
    None,
)?;
```

However, it's an experimental PR, so I'm not sure I can.

kyren:
This is really cool, thank you for putting work into this!  I'm sorry it's taken me a while to get around to looking at this, I had to focus on releasing 0.14.0 with a few important performance fixes first.

This is a great starting point for a 0.15 release though!  I have a couple of questions...

Are the `IntoTable` and `FromTable` traits necessary?  Can #[derive(LuaTable)] just output an implementation of ToLua / FromLua directly?

Related to the first question, is it considered bad style for a procedural derive to derive a trait using an identifier that's not the same name as the trait(s) that are implemented?  Do we need to do something like #[derive(LuaIntoTable, LuaFromTable)]?

What are the problems you see around integration with UserData?  To put it another way, I'm having a bit of trouble understanding your idea here:

```
#[derive(LuaTable, LuaProtoType, Debug, Default)]
struct Vec2D(f64, f64);
```

I assume `LuaProtoType` would implement the `Vec2D.new` method in Lua, but how can a type be both a `UserData` AND be convertible to/from a Lua table?

It WOULD be really neat to eventually have `__index` metamethod based field access on `UserData` types, though I'm not exactly sure how that would look with proc macros and still being able to implement `UserData` yourself.

This is all really good though!  I'm probably going to play with this this weekend and see how this all works.  I think I'd like to at the very least provide a procedural macro to do exactly what you've done here, convert based on field name from / into tables, in the very near future.  I think that future things like field access on `UserData` is much further off, so I'm interested to hear how you think that would integrate so that we can find a subset of this functionality that we can add in the near future without going in the wrong direction for the further off future.

kngwyu:
Thanks for feedback!

> Are the IntoTable and FromTable traits necessary? Can #[derive(LuaTable)] just output an implementation of ToLua / FromLua directly?

Yes. It's not necessary. We can implement `ToLua/FromLua` directly.

I implemented `IntoTable/FromTable` just because I want clarify what this macro does(define conversion between Rust's struct <=> Lua's table).

If you like `#derive[ToLua, FromLua]`, I'd modify immediately:wink:

> Related to the first question, is it considered bad style for a procedural derive to derive a trait using an identifier that's not the same name as the trait(s) that are implemented? Do we need to do something like #[derive(LuaIntoTable, LuaFromTable)]?

Yes, you're right. Though I thought `LuaTable` is more user-friendly, now I feel it's better to use same name as trait, following manner. 

> What are the problems you see around integration with UserData? 

Now, `UserData` automatically implements `ToLua/FromLua`, so we can't convert `T: UserData` into Lua's table when `to_lua` is called.

So, we have to implement 3 types of `ToLua/FromLua`,
- For 'T: UserData'
- For 'T: LuaTable'
- For 'T: LuaTable + UserData'

> I assume LuaProtoType would implement the Vec2D.new method in Lua, but how can a type be both a UserData AND be convertible to/from a Lua table?

In current implementation, If type T derives `LuaTable`, 
- we can convert variable typed T(e.g. `let a: T;`) into Lua's Table.
- we can convert Lua's variable which has keys with same names as T's members into Rust's variable.

And, in my idea, if type T derives `LuaProtoType`, we can set 
``` lua
T = {}
function T.new(x, ..) 
    local t = {}
    t.x = ~
    ...
    T.__index = T
    setmetatable(t, T)
    return t
end
```
in Lua. 

So, if we want to make a variable with type `T` both a `UserData` and `LuaTable`, we have to do in `ToLua` path
```lua
t = {}.
t.x = ~ -- set members
...
t.__index = T
```

However, I'm really not sure.

I have to read 'UserData' implementation more precisely  :pensive:

kyren:
Unfortunately, you can't really mix table conversion and `UserData` conversion, `UserData` is fundamentally based around holding the real rust type inside Lua as "userdata", and something in Lua can't be a userdata and a table at the same time.  The closest thing I can think of would be automatically converting back and forth between a table and some rust type before calling methods on it, but it would be very surprising, and very slow, to do that.

You *could* however provide what looked like field access to userdata in Lua by using the `__index` metamethod, and in fact you can already do this yourself with `rlua`, but it's not automatic of course.  This idea is what I was referring to above as a possible direction to go after a procedural derive for table conversion, but I'm not sure the two ideas even need to be related.

kngwyu:
Thanks for explanation.
I've read Programming Lua 4th edition today, and now I understand what `UserData` is, and sorry for my poor knowledge:pensive: 

> You could however provide what looked like field access to userdata in Lua by using the __index metamethod, and in fact you can already do this yourself with rlua, but it's not automatic of course. This idea is what I was referring to above as a possible direction to go after a procedural derive for table conversion, but I'm not sure the two ideas even need to be related.

Hmm...
'two ideas' are table-conversion and __index metamethod?
If we provide a class-like table based on prototype, table conversion might be useful, in the case like
```rust
#[derive(LuaTable, LuaProtoType)]
struct Point {
    x: i32,
    y: i32,
}
llua.globals().set_prototype::<Point>::();
lua.exec::<()>(
    r#"
p = Point.new(3, 4)
"#,
    None,
)?;
let p: Table = lua.globals().get("p").unwrap();
```

PR #79, opened 2018-05-14
find system lua with pkg-config

acrisci:
Hi,

I'm working on a project that requires to use the system lua and I'm having some problems building it because of linker errors.

The file `src/ffi.rs` adds the linker flag `-llua5.3` to the build, but the linker cannot find this on my system (I am using Fedora 28). The correct linker flags on my system are:

```
$ pkg-config --libs lua        
-llua -lm -ldl
```

So I removed the `#[link]` attribute from the ffi source file and it seems to still build fine (I think it gets automatically added in that build declaration in `build.rs` but I'm not sure) so the hardcoded `-llua5.3` linker argument does not get passed.

Then I added the pkg-config crate which adds all the correct flags for my system lua hopefully in a way that is portable.

I'm new to Rust, so I don't know if I did this correctly. Let me know if there is a different approach that will work better.

Thanks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ compiletest_rs = { version = "0.3", optional = true }
 
 [build-dependencies]
 cc = { version = "1.0", optional = true }
+pkg-config = "0.3.11"
 
 [dev-dependencies]
 rustyline = "1.0.0"
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,9 @@
 #[cfg(feature = "builtin-lua")]
 extern crate cc;
 
+#[cfg(not(feature = "builtin-lua"))]
+extern crate pkg_config;
+
 fn main() {
     #[cfg(feature = "builtin-lua")]
     {
@@ -61,5 +64,9 @@ fn main() {
             .file("lua/lvm.c")
             .file("lua/lzio.c")
             .compile("liblua5.3.a");
+    }
+    #[cfg(not(feature = "builtin-lua"))]
+    {
+        pkg_config::Config::new().atleast_version("5.3").probe("lua").unwrap();
     }
 }
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -78,7 +78,6 @@ pub const LUA_GCSETPAUSE: c_int = 6;
 pub const LUA_GCSETSTEPMUL: c_int = 7;
 pub const LUA_GCISRUNNING: c_int = 9;
 
-#[link(name = "lua5.3")]
 extern "C" {
     pub fn lua_newstate(alloc: lua_Alloc, ud: *mut c_void) -> *mut lua_State;
     pub fn lua_close(state: *mut lua_State);


kyren:
Thanks for the work on this!  Before I merge this, I have a bunch of questions.  I'd actually like to use this as an excuse to talk through a bunch of linker things (also with @Timidger and @psychon) because I'm not sure what's appropriate here and there are a lot of related issues.

What adds the linker flags for lua5.3, `pkg_config::Config::probe` seems to return a `pkg_config::Library`, does it also print to stdout to add the linker flags?

Are libraries compiled with the `cc` crate automatically linked, implying that you don't need the `link` attribute on the `extern "C"` block?

Should we do something if pkg-config doesn't find lua 5.3, like try linking to a hard-coded `lua5.3` as a fallback?

Ultimately, we probably need to know facts about how the system lua was built which only exist inside the luaconf.h header shipped with the compiled library.  What is the proper way of solving this?  Should I be using bindgen or something similar to ensure that constants are the correct value?  Previously I was using the `builtin-lua` feature as a hack to do something very chucklefish specific, if other people are disabling it I'd like to make it actually correct in the face of things like changing the type of lua_Number or changing other configurable constants.  Currently this will just lead to potential UB later, so this is kind of a big deal.

Should the `builtin-lua` feature even exist?  I'm not 100% positive that a cargo feature is the right way to decide on build settings, I've gotten the impression before that this might be solved better with something like an environment variable switch, but I'm not super sure what's the best thing to do here.  I think building a predictable embedded Lua is *probably* the right default, but maybe not if we can determine enough facts about the system lua version?

`rlua` currently has a git `failure` dependency, because I needed to be able to disable backtraces and was expecting `failure` 1.0 to release in mid-march.  If way-cooler needs this fixed, I guess they would want a new `rlua` release soon?  I was waiting on `failure` 1.0 to do the next version, actually, do you need me to back out of the `failure` version change?

Lua 5.4 might be coming out soon, and compatibility between Lua 5.3 and 5.4 should be very easy to accomplish.  What is the *correct* mechanism to allow the user of `rlua` to switch between 5.3 and 5.4, is it cargo features?  I include this in case it affects the answers to any other questions.

Sorry to dump all this on what is otherwise a pretty simple PR, but some of these have been nagging at me for a while because I don't know what's the best pattern to follow.

Timidger:
> What adds the linker flags for lua5.3, pkg_config::Config::probe seems to return a pkg_config::Library, does it also print to stdout to add the linker flags?

Yes

> When using the bulitin-lua feature, are libraries compiled with the cc crate automatically linked, implying that you don't need the link attribute on the extern "C" block?

No, make sure you emit the correct flags for built in behavior. [see this](https://doc.rust-lang.org/cargo/reference/buildscripts.html). 

> Should we do something if pkg-config doesn't find lua 5.3, like try linking to a hard-coded lua5.3 as a fallback?

If you wish, that's up to you. Probably better to just fail because AFAIK pkg config should always work. If there's some environmental it doesn't that can be patched later.

>... Currently this will just lead to potential UB later, so this is kind of a big deal

Yes that would be prudent. If you open up some issues detailing what needs to be done I can look into doing them if you're too busy to get to it soon. I take soundness problems in Way Cooler very seriously.

> Should the builtin-lua feature even exist?

Yes, this is the correct way to do things in my opinion. 

>  I was waiting on failure 1.0 to do the next version, actually, do you need me to back out of the failure version change?

No that is fine. I think I'm a few releases behind to be honest. 

> What is the correct mechanism to allow the user of rlua to switch between 5.3 and 5.4, is it cargo features?

That's tricky, I'm not sure. You might want to look into multiple options for that.

acrisci:
>> When using the bulitin-lua feature, are libraries compiled with the cc crate automatically linked, implying that you don't need the link attribute on the extern "C" block?

> No, make sure you emit the correct flags for built in behavior. see this.

I'm going to call this into question.

I made an example project with a dependency on this branch with:

```
rlua = { path = "../rlua", default-features = true }
```

And `cargo build --verbose` gives this in the logs:

```
     Running `rustc --crate-name rlua /home/tony/projects/rlua/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 --cfg 'feature="builtin-lua"' --cfg 'feature="cc"' --cfg 'feature="default"' -C metadata=e2eeaeec6412b91c -C extra-filename=-e2eeaeec6412b91c --out-dir /home/tony/projects/rust-test/target/debug/deps -C incremental=/home/tony/projects/rust-test/target/debug/incremental -L dependency=/home/tony/projects/rust-test/target/debug/deps --extern libc=/home/tony/projects/rust-test/target/debug/deps/liblibc-fa1af275a382d06f.rlib --extern failure=/home/tony/projects/rust-test/target/debug/deps/libfailure-de7d7927d832d199.rlib -L native=/home/tony/projects/rust-test/target/debug/build/rlua-2ab3a82c3f56c3dc/out -l static=lua5.3`
```

Notice `-l static=lua5.3` and the end. I don't know if it's coming from "cc" or what, but it's there without that link build declaration thing in `ffi.c` and no cargo println! things in the build.rs.

Then with:

```
rlua = { path = "../rlua", default-features = false }
```

I get:

```
     Running `rustc --crate-name rlua /home/tony/projects/rlua/src/lib.rs --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=69cdecf5f833c1b6 -C extra-filename=-69cdecf5f833c1b6 --out-dir /home/tony/projects/rust-test/target/debug/deps -C incremental=/home/tony/projects/rust-test/target/debug/incremental -L dependency=/home/tony/projects/rust-test/target/debug/deps --extern libc=/home/tony/projects/rust-test/target/debug/deps/liblibc-fa1af275a382d06f.rlib --extern failure=/home/tony/projects/rust-test/target/debug/deps/libfailure-de7d7927d832d199.rlib -l lua -l m -l dl`
```

And you'll notice at the end it gives the pkg-config flags in the first post.

I think this indicates that the PR will give the correct linker flags how it is now, but someone should verify this.

Timidger:
Ah seems I'm wrong then, my mistake! Thanks for checking me @acrsci

jonas-schievink:
Yeah, `cc` does this by default: https://docs.rs/cc/1.0.15/cc/struct.Build.html#method.cargo_metadata

> Define whether metadata should be emitted for cargo allowing it to automatically link the binary. Defaults to true.

kyren:
> If you wish, that's up to you. Probably better to just fail because AFAIK pkg config should always work. If there's some environmental it doesn't that can be patched later.

Does this include windows?  I suppose on windows you wouldn't be able to find a lua library reliably at all, so maybe the appropriate thing to do is NOT call `unwrap` on the result, and then let the user of rlua manually link in lua from elsewhere?  What I don't want to happen is the lack of `pkg-config` to cause a panic if you aren't using `builtin-lua`.

> Yes that would be prudent. If you open up some issues detailing what needs to be done I can look into doing them if you're too busy to get to it soon. I take soundness problems in Way Cooler very seriously.

That's kind of the problem, I don't know what needs to be done at all.  You have to find a header file and parse configuration out of it, so I really don't know the best way to do this.  Perhaps the FFI bindings should just be generated by bindgen entirely, and that's the best way to do this?

In the meantime I think this patch can be merged as long as it doesn't cause any damage, so really the only thing I think needs to be done is to make sure that disabling `builtin-lua` doesn't cause an unavoidable build error on windows, though I seriously doubt anybody was using it like that anyway.

The very next thing we should do though is start reading values from `luaconf.h` somehow.

acrisci:
> The very next thing we should do though is start reading values from luaconf.h somehow.

Just going to throw out a suggestion. Maybe compile a test program as part of the build process that probes the library and looks for features?

jonas-schievink:
Is it really that important to check `luaconf.h`? I realize that an incompatible configuration might lead to UB, but then again, when you're trying to link to the system Lua, you can be pretty sure that it uses the default configuration since otherwise many libraries and scripts won't work with it.

If all requirements are properly documented, the few people who want to link to their own custom non-standard Lua can just ensure those are met. AFAIK, those are (most are mentioned in [Cargo.toml](https://github.com/chucklefish/rlua/blob/60bfe1fb077a75b37fa2aef1883db45298761987/Cargo.toml#L21-L24) already):

* Integers are `i64`
* Floats are `f64`
* Max stack size is 1,000,000 (this is the default with 64 bit integers anyway)
* The extra space in Lua states (`LUA_EXTRASPACE`) is at least a pointer in size (and has pointer alignment - Lua doesn't explicitly mention this, but as long as the Lua state contains at least one pointer it should be the case)
* Disabling coercions will change behaviour in `coerce_*` methods (at least I think so)

> Should we do something if pkg-config doesn't find lua 5.3, like try linking to a hard-coded lua5.3 as a fallback?

Unless I'm misunderstanding how pkg-config works, throwing an error without a way to specify how Lua should be linked would effectively preclude linking custom Luas altogether. You can either use rlua's builtin Lua with its stock `luaconf.h`, or link against the system Lua (which also has a mostly stock `luaconf.h`).

Timidger:
> Does this include windows? I suppose on windows you wouldn't be able to find a lua library reliably at all, so maybe the appropriate thing to do is NOT call unwrap on the result, and then let the user of rlua manually link in lua from elsewhere? What I don't want to happen is the lack of pkg-config to cause a panic if you aren't using builtin-lua.

Oh right Windows... I always forget about Windows 

> Perhaps the FFI bindings should just be generated by bindgen entirely, and that's the best way to do this?

Yes that's probably the best way to do this.

psychon:
> You have to find a [Lua] header file and parse configuration out of it, so I really don't know the best way to do this. Perhaps the FFI bindings should just be generated by bindgen entirely, and that's the best way to do this?

Side note: More often than I'd like, I see bug reports about problems building awesome. The problem is that e.g. Lua headers from Lua 5.2 and liblua5.3.so are mixed (or things like that). So, somehow finding a file called `luaconf.h` and using the information from there is not fool-proof in a world where distros install all major Lua versions side-by-side. (But this is partly offtopic, since rlua only works with lua 5.3 and distro use the default settings anyway (right?))

kyren:
Sorry for letting this languish for so long, after thinking about this some more I think ONE change would make this a strict improvement over what's there now.

Would it be possible to have another feature for using `pkg-config`?  What I'm thinking is, since I actually have used rlua this way, it would be nice to be able to just build rlua and attempt no linking at all and let the user link in an appropriate Lua instance.  This is very useful when you have to patch Lua or build Lua for esoteric environments, and useful for windows since pkg-config rarely works there.

Since `builtin-lua` is the default, what I'm thinking is just do exactly this change except make another feature (call it, I dunno, `system-lua`?) that depends on `pkg-config`, make `pkg-config` optional, and only perform the `pkg_config::Config` step if `system-lua` is enabled.  Since you already need to mess with features to NOT use `builtin-lua`, this is not such a huge hardship.

Without this change, `pkg-config` is always used if `builtin-lua` is disabled, and since it's a fatal error if `pkg-config` doesn't find lua53, it doesn't work well when trying to build for weird environments.

Instead of making a new feature, you could just use `pkg-config` itself as the feature, but I don't think this meshes well with other dependencies that might use `pkg-config` themselves or it being unintentionally enabled in other ways?  I'm a bit fuzzy atm on how this precisely works in cargo.

kyren:
I'm going to go ahead and merge this and incorporate my proposed changes, since hopefully that solves both of our sets of problems.

Issue #85, opened 2018-09-03
Collect common problems into FAQ


poga:
While using `rlua`, I encountered a lot of lifetime problem which are already described in previous issues (e.g. https://github.com/kyren/rlua/issues/20#issuecomment-405751583, https://github.com/kyren/rlua/issues/41, https://github.com/kyren/rlua/issues/74...). After reading those issues, using `rlua` becomes much easier.

It would be great if we can collect these design rationale into a FAQ document for beginners to lookup. I can help if you're ok about the idea.

Here's a list of issues/comment I think are really helpful:

#### Why `'static` lifetime is required when passing data to `Lua`?

> The reason I mention the GC is because that is the reason WHY userdata must be 'static: since we don't know when Lua will get around to garbage collecting our userdata, we can't make any assumptions about the lifetimes of any references in the userdata. Lua::scope forces the userdatas to be destructed at the exit of the scope.

* https://github.com/kyren/rlua/issues/74#issuecomment-370152884
* https://github.com/kyren/rlua/issues/20#issuecomment-405751583

#### How do I store `Table`, `Function`, or other `rlua` reference types outside of Lua? 

> There are two rules about rlua reference types that you're butting up against:
> 
> rlua reference types (Table, Function, etc) hold references to Lua, and are not really designed to be stored along-side Lua, because that would require self borrows. You CAN make structs in Rust that self borrow, but you really don't want to go down that road, it's very complex and you 99.999% of the time don't need it.
>
> rlua reference types are also not designed to be stored inside userdata inside Lua. When I say userdata here, I mean both actual UserData types and also things like Rust callbacks. For one, there are some actual safety issues if you were able to do that, but more importantly Lua itself is not really designed for this either. Lua's C API does not provide a way of telling Lua that a userdata contains a registry reference, so you would end up potentially making something that cannot be garbage collected. The actual issue is very complicated but you can read about it some here.

* https://github.com/kyren/rlua/issues/73#issuecomment-370222198

kyren:
Yeah, very very much agreed.  I'm definitely okay with the idea if you want to take a stab at it, and I might have some more things in the FAQ after you're done as well.

Some of the discussion in the comments you linked may not apply in the near future though if we merge #86, but then that will just make NEW things that need to go in the FAQ.

zacps:
I'm not sure if this counts as a common problem, but I couldn't figure out how to set the include path for `require`. Might just be my lack of knowledge of lua showing though.

poga:
[...]

PR #87, opened 2018-09-05
add FAQ

poga:
As #85 discussed, I tried to collect some of the common problem while using `rlua` into a FAQ.

For now, it's pretty basic. It's just a collection of helpful links and discussions.


jugglerchris:
Merged (finally) with some tweaks - thank you!

PR #92, opened 2018-09-27
Adds missing cause when formatting CallbackError.

shawnscode:


kyren:
I did this because the generated Format impls in the `failure` crate don't seem to print the cause chain either, I was under the impression that you're not supposed to do that, and instead print the cause chain manually when the error is handled.

If every implementation of `Failure` did this, it would be fine to print the cause as part of `Display` because that would print the whole chain.  However, all implementations of `Failure` *don't* do this, so you have to at some point walk the cause chain and print each cause manually, at which point any error that prints its cause prints duplicates.

I was assuming at some point that there would be a sanctioned way in `failure` of printing the cause chain.  At some point I ended up making a [generic error wrapper](https://gist.github.com/kyren/a661e91caddba84044a048ad3b0adc22) that walked the cause chain in its Display impl and printed the error *and* any available backtraces, but something like this should probably exist externally?

In any case, I'm not sure that printing the cause in the Display impl is right and purposefully avoided it, but I really don't know what is best practice here, and I admit it's pretty inconvenient as it is.  I found [an issue](https://github.com/rust-lang-nursery/failure/issues/153) on the failure repo discussing the problem.

kyren:
I've used rlua inside a large project, where I would encounter errors with several layers of Rust and Lua on the stack, and that's the time where I adjusted the error Display impl to avoid repeating information.  I don't know if it's still correct, though, so rather than talking about it abstractly let's come up with a concrete example.

I'll come up with a concrete example of an error where the stack looks like "Rust -> Lua -> Rust -> Lua" and we'll see what it looks like with different approaches.

Issue #81, opened 2018-07-07
Support for lua_sethook


beatgammit:
This might be a safety nightmare, so I wanted to get your input before looking into implementation. Here are the production use-cases:

* running untrusted code submitted from users (stop infinite loops)
* pre-empting a CPU-heavy lua script in an async event loop

And then there's the typical debugging tasks, but I think those can be done within lua with the current debug library.

From #52, it looks like the debug library is considered "unsafe", and I'm not sure if this can be done safely. In any case, I'd really like be able to handle the above use cases.

I created a similar issue on hlua ([#194](https://github.com/tomaka/hlua/issues/194)), just in case discussion in the other project is useful here.

I'm willing to do work on this, but I'd like to know your thoughts on feasibility first.

EDIT: This is essentially the first bullet point in the README (and perhaps the second).

kyren:
This is now added!

Running untrusted code using PUC-Rio Lua (by itself or via `rlua`, in any way really) is risky in a lot of ways.  I added some information to the README about this so that the sandboxing pieces of the `rlua` API did not become an "attractive nuisance".

PR #95, opened 2018-10-08
Support for hooks (viewing only)

lemarcuspoilus:
This PR adds three new functions and a few other types to safely wrap the `lua_sethook` function, as suggested by the issue https://github.com/kyren/rlua/issues/81.

This is my best attempt to make it possible to consult information for debugging purposes and to control script execution.

These functions have been written:
- `set_hook` and `set_mut_hook`: Sets a function to be called when the hook is raised. It gets the new `Debug` struct containing information related to the code and function (depending on the hook mask)
- `remove_hook`: Removes the hook previously set

A few new types such as `Debug` and `HookOptions`, written inside a new module, are used as parameters to those functions above. That module also contains a procedure given to `lua_sethook` used to call the Rust callback safely.

I regard this change as safe since it does not expose functions that changes the Lua stack or anything else (other than using the `Lua` structure itself). Moreover, since the hook can return a `Result`, it can terminate the execution of a script, as shown in a test, by returning the error back to the caller of `exec` and such. This satisfies the desire of the author as noted in the README.

Anyone is welcome to comment or suggest changes to make on my PR, since I did it without knowing too much about Lua's internal workings. I have not changed the README or version; I leave this to rlua's author discretion.

Thank you.

### DIFF
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -152,14 +152,15 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             _unused: ()
         };
 
-        let cb = (&*extra_data(state)).hook_callback
+        let cb = (extra_data(state)).hook_callback
             .as_ref()
+            .map(|rc| rc.clone())
             .expect("rlua internal error: no hooks previously set; this is a bug");
         (&mut *match cb.try_borrow_mut() {
             Ok(b) => b,
             Err(_) => rlua_panic!("Lua should not allow hooks to be called within another hook;\
                 please make an issue")
-        })(lua, &debug)
+        })(&lua, &debug)
     });
 }
 
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -6,6 +6,12 @@ use lua::extra_data;
 use util::callback_error;
 
 /// Contains information about the running code at the moments specified when setting the hook.
+/// Each field is described in the [Lua 5.3 documentaton][lua_doc].
+///
+/// Depending on when the hook is called, some fields might not be available. In these cases,
+/// integers and booleans might not be valid and/or strings are set to `None`.
+///
+/// [lua_doc]: https://www.lua.org/manual/5.3/manual.html#lua_Debug
 #[derive(Clone, Debug)]
 pub struct Debug<'a> {
     pub name: Option<Cow<'a, str>>,
@@ -23,7 +29,7 @@ pub struct Debug<'a> {
 }
 
 impl<'a> Debug<'a> {
-    /// Constructs a new `Debug` structure that is not associated with a Lua debug structure. It
+    /// Construct a new `Debug` structure that is not associated with a Lua debug structure. It
     /// involves some string copying.
     pub fn to_owned(&'a self) -> Debug<'static> {
         Debug {
@@ -43,7 +49,7 @@ impl<'a> Debug<'a> {
     }
 }
 
-/// Indicates in which circumstances the hook should be called by Lua.
+/// Indicate in which circumstances the hook should be called by Lua.
 pub struct HookOptions {
     /// Before a function call.
     pub calls: bool,
@@ -53,13 +59,13 @@ pub struct HookOptions {
     pub lines: bool,
     /// After a certain amount of instructions specified by `count`.
     pub after_counts: bool,
-    /// Indicates how many instructions to execute before calling the hook. Only effective when
+    /// Specify how many instructions to execute before calling the hook. Only effective when
     /// `after_counts` is set to true.
     pub count: u32
 }
 
 impl HookOptions {
-    // Computes the mask to pass to `lua_sethook`.
+    // Compute the mask to pass to `lua_sethook`.
     pub(crate) fn mask(&self) -> c_int {
         let mut mask: c_int = 0;
         if self.calls { mask |= ffi::LUA_MASKCALL }
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -133,8 +133,6 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             rlua_panic!("lua_getinfo failed")
         }
 
-        let extra = &mut *extra_data(state);
-
         let debug = Debug {
             name: ptr_to_str((*ar).name as *const i8),
             namewhat: ptr_to_str((*ar).namewhat as *const i8),
@@ -153,10 +151,10 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             _unused: ()
         };
 
-        let cb = extra.hook_callback
-            .as_mut()
+        let cb = (&*extra_data(state)).hook_callback
+            .as_ref()
             .expect("rlua internal error: no hooks previously set; this is a bug");
-        cb(&debug)
+        (*cb)(&debug)
     });
 }
 
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -70,7 +70,7 @@ impl<'a> Debug<'a> {
 /// # use rlua::HookTriggers;
 /// # fn main() {
 /// let triggers = HookTriggers {
-///     after_counts: Some(1), ..Default::default()
+///     on_every_nth_instruction: Some(1), ..Default::default()
 /// };
 /// # let _ = triggers;
 /// # }
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -77,41 +77,41 @@ impl<'a> Debug<'a> {
 /// ```
 pub struct HookTriggers {
     /// Before a function call.
-    pub calls: bool,
+    pub on_calls: bool,
     /// When Lua returns from a function.
-    pub returns: bool,
+    pub on_returns: bool,
     /// Before executing a new line, or returning from a function call.
-    pub lines: bool,
+    pub each_line: bool,
     /// After a certain amount of instructions. When set to `Some(count)`, `count` is the number of
     /// instructions to execute before calling the hook.
-    pub after_counts: Option<u32>,
+    pub on_every_nth_instruction: Option<u32>,
 }
 
 impl HookTriggers {
     /// Compute the mask to pass to `lua_sethook`.
     pub(crate) fn mask(&self) -> c_int {
         let mut mask: c_int = 0;
-        if self.calls { mask |= ffi::LUA_MASKCALL }
-        if self.returns { mask |= ffi::LUA_MASKRET }
-        if self.lines { mask |= ffi::LUA_MASKLINE }
-        if self.after_counts.is_some() { mask |= ffi::LUA_MASKCOUNT }
+        if self.on_calls { mask |= ffi::LUA_MASKCALL }
+        if self.on_returns { mask |= ffi::LUA_MASKRET }
+        if self.each_line { mask |= ffi::LUA_MASKLINE }
+        if self.on_every_nth_instruction.is_some() { mask |= ffi::LUA_MASKCOUNT }
         mask
     }
 
     /// Returns the `count` parameter to pass to `lua_sethook`, if applicable. Otherwise, zero is
     /// returned.
     pub(crate) fn count(&self) -> c_int {
-        self.after_counts.unwrap_or(0) as c_int
+        self.on_every_nth_instruction.unwrap_or(0) as c_int
     }
 }
 
 impl Default for HookTriggers {
     fn default() -> Self {
         HookTriggers {
-            calls: false,
-            returns: false,
-            lines: false,
-            after_counts: None
+            on_calls: false,
+            on_returns: false,
+            each_line: false,
+            on_every_nth_instruction: None
         }
     }
 }
@@ -145,7 +145,7 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
 
         let cb = extra.hook_callback
             .as_mut()
-            .expect("no hooks previously set; this is a bug");
+            .expect("rlua internal error: no hooks previously set; this is a bug");
         cb(&debug)
     });
 }
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -1,11 +1,24 @@
-use libc::c_int;
+use std::{slice, ptr, mem, str};
+use libc::{self, c_int};
 use ffi::{self, lua_State, lua_Debug};
 use lua::extra_data;
 use util::callback_error;
 
 /// Contains information about the running code at the moments specified when setting the hook.
-pub struct Debug {
-    pub curr_line: u32
+#[derive(Clone, Debug)]
+pub struct Debug<'a> {
+    pub name: Option<&'a str>,
+    pub namewhat: Option<&'a str>,
+    pub what: Option<&'a str>,
+    pub source: Option<&'a str>,
+    pub curr_line: u32,
+    pub line_defined: u32,
+    pub last_line_defined: u32,
+    pub num_ups: u32,
+    pub num_params: u32,
+    pub is_vararg: bool,
+    pub is_tailcall: bool,
+    pub short_src: Option<&'a str>
 }
 
 /// Indicates in which circumstances the hook should be called by Lua.
@@ -50,10 +63,27 @@ impl Default for HookOptions {
 /// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
 pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
     callback_error(state, || {
+        if ffi::lua_getinfo(state, cstr!("nSltu"), ar) == 0 {
+            rlua_panic!("lua_getinfo failed")
+        }
+
         let extra = &mut *extra_data(state);
 
         let debug = Debug {
-            curr_line: (*ar).currentline as u32
+            name: ptr_to_str((*ar).name as *const i8),
+            namewhat: ptr_to_str((*ar).namewhat as *const i8),
+            what: ptr_to_str((*ar).what as *const i8),
+            source: ptr_to_str((*ar).source as *const i8),
+            curr_line: (*ar).currentline as u32,
+            line_defined: (*ar).linedefined as u32,
+            last_line_defined: (*ar).lastlinedefined as u32,
+            num_ups: (*ar).nups as u32,
+            num_params: (*ar).nparams as u32,
+            is_vararg: (*ar).isvararg == 1,
+            is_tailcall: (*ar).istailcall == 1,
+            short_src: str::from_utf8(mem::transmute((*ar).short_src.as_ref()))
+                .and_then(|r| Ok(Some(r)))
+                .unwrap_or(None)
         };
 
         let cb = extra.hook_callback
@@ -61,4 +91,13 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             .expect("no hooks previously set; this is a bug");
         cb(&debug)
     });
+}
+
+unsafe fn ptr_to_str<'a>(input: *const i8) -> Option<&'a str> {
+    if input == ptr::null() || ptr::read(input) == 0 {
+        return None;
+    }
+    let len = libc::strlen(input) as usize;
+    let input = slice::from_raw_parts(input as *const u8, len);
+    str::from_utf8(input).and_then(|r| Ok(Some(r.trim_right()))).unwrap_or(None)
 }
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -1,5 +1,6 @@
 use std::{slice, ptr, mem, str};
 use std::borrow::Cow;
+use std::sync::atomic::Ordering;
 use libc::{self, c_int};
 use ffi::{self, lua_State, lua_Debug};
 use lua::extra_data;
@@ -128,7 +129,7 @@ impl Default for HookTriggers {
 
 /// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
 pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
-    callback_error(state, || {
+    let result = callback_error(state, || {
         if ffi::lua_getinfo(state, cstr!("nSltu"), ar) == 0 {
             rlua_panic!("lua_getinfo failed")
         }
@@ -151,11 +152,14 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             _unused: ()
         };
 
+        (&mut *extra_data(state)).in_hook.store(true, Ordering::Relaxed);
         let cb = (&*extra_data(state)).hook_callback
             .as_ref()
             .expect("rlua internal error: no hooks previously set; this is a bug");
         (*cb)(&debug)
     });
+    (&mut *extra_data(state)).in_hook.store(false, Ordering::Relaxed);
+    result
 }
 
 unsafe fn ptr_to_str<'a>(input: *const i8) -> Option<Cow<'a, str>> {
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -1,6 +1,5 @@
 use std::{slice, ptr, mem, str};
 use std::borrow::Cow;
-use std::sync::atomic::Ordering;
 use libc::{self, c_int};
 use ffi::{self, lua_State, lua_Debug};
 use lua::extra_data;
@@ -129,7 +128,7 @@ impl Default for HookTriggers {
 
 /// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
 pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
-    let result = callback_error(state, || {
+    callback_error(state, || {
         if ffi::lua_getinfo(state, cstr!("nSltu"), ar) == 0 {
             rlua_panic!("lua_getinfo failed")
         }
@@ -152,14 +151,11 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             _unused: ()
         };
 
-        (&mut *extra_data(state)).in_hook.store(true, Ordering::Relaxed);
         let cb = (&*extra_data(state)).hook_callback
             .as_ref()
             .expect("rlua internal error: no hooks previously set; this is a bug");
         (*cb)(&debug)
     });
-    (&mut *extra_data(state)).in_hook.store(false, Ordering::Relaxed);
-    result
 }
 
 unsafe fn ptr_to_str<'a>(input: *const i8) -> Option<Cow<'a, str>> {
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -1,0 +1,64 @@
+use libc::c_int;
+use ffi::{self, lua_State, lua_Debug};
+use lua::extra_data;
+use util::callback_error;
+
+/// Contains information about the running code at the moments specified when setting the hook.
+pub struct Debug {
+    pub curr_line: u32
+}
+
+/// Indicates in which circumstances the hook should be called by Lua.
+pub struct HookOptions {
+    /// Before a function call.
+    pub calls: bool,
+    /// When Lua returns from a function.
+    pub returns: bool,
+    /// Before executing a new line, or returning from a function call.
+    pub lines: bool,
+    /// After a certain amount of instructions specified by `count`.
+    pub after_counts: bool,
+    /// Indicates how many instructions to execute before calling the hook. Only effective when
+    /// `after_counts` is set to true.
+    pub count: u32
+}
+
+impl HookOptions {
+    // Computes the mask to pass to `lua_sethook`.
+    pub(crate) fn mask(&self) -> c_int {
+        let mut mask: c_int = 0;
+        if self.calls { mask |= ffi::LUA_MASKCALL }
+        if self.returns { mask |= ffi::LUA_MASKRET }
+        if self.lines { mask |= ffi::LUA_MASKLINE }
+        if self.after_counts { mask |= ffi::LUA_MASKCOUNT }
+        mask
+    }
+}
+
+impl Default for HookOptions {
+    fn default() -> Self {
+        HookOptions {
+            calls: false,
+            returns: false,
+            lines: false,
+            after_counts: false,
+            count: 0
+        }
+    }
+}
+
+// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
+pub unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
+    callback_error(state, || {
+        let extra = &*extra_data(state);
+
+        let debug = Debug {
+            curr_line: (*ar).currentline as u32
+        };
+
+        let cb = extra.hook_callback
+            .as_ref()
+            .expect("no hooks previously set; this is a bug");
+        cb(&debug)
+    });
+}
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -133,7 +133,7 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             rlua_panic!("lua_getinfo failed")
         }
 
-        let lua = Lua::from_state(state, false);
+        let lua = Lua::from_state(state);
         let debug = Debug {
             name: ptr_to_str((*ar).name as *const i8),
             namewhat: ptr_to_str((*ar).namewhat as *const i8),
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -147,6 +147,7 @@ pub struct Stack {
 /// # let _ = triggers;
 /// # }
 /// ```
+#[derive(Clone, Display)]
 pub struct HookTriggers {
     /// Before a function call.
     pub on_calls: bool,
@@ -184,17 +185,6 @@ impl HookTriggers {
     /// returned.
     pub(crate) fn count(&self) -> c_int {
         self.every_nth_instruction.unwrap_or(0) as c_int
-    }
-}
-
-impl Default for HookTriggers {
-    fn default() -> Self {
-        HookTriggers {
-            on_calls: false,
-            on_returns: false,
-            every_line: false,
-            every_nth_instruction: None
-        }
     }
 }
 
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -47,17 +47,17 @@ impl Default for HookOptions {
     }
 }
 
-// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
-pub unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
+/// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
+pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
     callback_error(state, || {
-        let extra = &*extra_data(state);
+        let extra = &mut *extra_data(state);
 
         let debug = Debug {
             curr_line: (*ar).currentline as u32
         };
 
         let cb = extra.hook_callback
-            .as_ref()
+            .as_mut()
             .expect("no hooks previously set; this is a bug");
         cb(&debug)
     });
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -133,7 +133,7 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             rlua_panic!("lua_getinfo failed")
         }
 
-        let lua = Lua::from_state(state);
+        let lua = Lua::make_ephemeral(state);
         let debug = Debug {
             name: ptr_to_str((*ar).name as *const i8),
             namewhat: ptr_to_str((*ar).namewhat as *const i8),
@@ -152,15 +152,16 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             _unused: ()
         };
 
-        let cb = (extra_data(state)).hook_callback
+        let cb = (&*extra_data(state)).hook_callback
             .as_ref()
             .map(|rc| rc.clone())
             .expect("rlua internal error: no hooks previously set; this is a bug");
-        (&mut *match cb.try_borrow_mut() {
-            Ok(b) => b,
+        let outcome = match cb.try_borrow_mut() {
+            Ok(mut b) => (&mut *b)(&lua, &debug),
             Err(_) => rlua_panic!("Lua should not allow hooks to be called within another hook;\
                 please make an issue")
-        })(&lua, &debug)
+        };
+        outcome
     });
 }
 
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -70,7 +70,7 @@ impl<'a> Debug<'a> {
 /// # use rlua::HookTriggers;
 /// # fn main() {
 /// let triggers = HookTriggers {
-///     on_every_nth_instruction: Some(1), ..Default::default()
+///     every_nth_instruction: Some(1), ..Default::default()
 /// };
 /// # let _ = triggers;
 /// # }
@@ -81,10 +81,20 @@ pub struct HookTriggers {
     /// When Lua returns from a function.
     pub on_returns: bool,
     /// Before executing a new line, or returning from a function call.
-    pub each_line: bool,
+    pub every_line: bool,
     /// After a certain amount of instructions. When set to `Some(count)`, `count` is the number of
     /// instructions to execute before calling the hook.
-    pub on_every_nth_instruction: Option<u32>,
+    ///
+    /// # Performance
+    ///
+    /// Setting this to a low value will certainly result in a large overhead due to the crate's
+    /// safety layer and convenience wrapped over Lua's low-level hooks. `1` is such an example of
+    /// a high overhead choice.
+    ///
+    /// Please find a number that is high enough so it's not that bad of an issue, while still
+    /// having enough precision for your needs. Keep in mind instructions are additions, calls to
+    /// functions, assignments to variables, etc.; they are very short.
+    pub every_nth_instruction: Option<u32>,
 }
 
 impl HookTriggers {
@@ -93,15 +103,15 @@ impl HookTriggers {
         let mut mask: c_int = 0;
         if self.on_calls { mask |= ffi::LUA_MASKCALL }
         if self.on_returns { mask |= ffi::LUA_MASKRET }
-        if self.each_line { mask |= ffi::LUA_MASKLINE }
-        if self.on_every_nth_instruction.is_some() { mask |= ffi::LUA_MASKCOUNT }
+        if self.every_line { mask |= ffi::LUA_MASKLINE }
+        if self.every_nth_instruction.is_some() { mask |= ffi::LUA_MASKCOUNT }
         mask
     }
 
     /// Returns the `count` parameter to pass to `lua_sethook`, if applicable. Otherwise, zero is
     /// returned.
     pub(crate) fn count(&self) -> c_int {
-        self.on_every_nth_instruction.unwrap_or(0) as c_int
+        self.every_nth_instruction.unwrap_or(0) as c_int
     }
 }
 
@@ -110,8 +120,8 @@ impl Default for HookTriggers {
         HookTriggers {
             on_calls: false,
             on_returns: false,
-            each_line: false,
-            on_every_nth_instruction: None
+            every_line: false,
+            every_nth_instruction: None
         }
     }
 }
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -2,7 +2,7 @@ use std::{slice, ptr, mem, str};
 use std::borrow::Cow;
 use libc::{self, c_int};
 use ffi::{self, lua_State, lua_Debug};
-use lua::extra_data;
+use lua::{Lua, extra_data};
 use util::callback_error;
 
 /// Contains information about the running code at the moments specified when setting the hook.
@@ -133,6 +133,7 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
             rlua_panic!("lua_getinfo failed")
         }
 
+        let lua = Lua::from_state(state, false);
         let debug = Debug {
             name: ptr_to_str((*ar).name as *const i8),
             namewhat: ptr_to_str((*ar).namewhat as *const i8),
@@ -154,7 +155,11 @@ pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_De
         let cb = (&*extra_data(state)).hook_callback
             .as_ref()
             .expect("rlua internal error: no hooks previously set; this is a bug");
-        (*cb)(&debug)
+        (&mut *match cb.try_borrow_mut() {
+            Ok(b) => b,
+            Err(_) => rlua_panic!("Lua should not allow hooks to be called within another hook;\
+                please make an issue")
+        })(lua, &debug)
     });
 }
 
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -1,56 +1,128 @@
-use std::{slice, ptr, mem, str};
-use std::borrow::Cow;
+use std::{slice, ptr, str};
+use std::marker::PhantomData;
 use libc::{self, c_int};
 use ffi::{self, lua_State, lua_Debug};
 use lua::{Lua, extra_data};
 use util::callback_error;
 
 /// Contains information about the running code at the moments specified when setting the hook.
-/// Each field is described in the [Lua 5.3 documentaton][lua_doc].
+/// All the documentation can be read in the [Lua 5.3 documentaton][lua_doc].
 ///
-/// Depending on when the hook is called, some fields might not be available. In these cases,
-/// integers and booleans might not be valid and/or strings are set to `None`.
+/// # Usage
+///
+/// For optimal performance, the user must call the methods that return the information they need to
+/// use. Those methods return a short structure with the data in the fields. That being said,
+/// depending on where the hook is called, some methods might return zeroes and empty strings or
+/// possibly bad values.
+///
+/// This structure does not cache the data obtained; you should only call each function once and
+/// keep the reference obtained for as long as needed. However, you may not outlive the hook
+/// callback; if you need to, please clone the internals that you need and move them outside.
+///
+/// # Panics
+///
+/// This structure contains methods that will panic if there is an internal error. If this happens
+/// to you, please make an issue to rlua's repository. as this behavior isn't normal.
 ///
 /// [lua_doc]: https://www.lua.org/manual/5.3/manual.html#lua_Debug
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Debug<'a> {
-    pub name: Option<Cow<'a, str>>,
-    pub namewhat: Option<Cow<'a, str>>,
-    pub what: Option<Cow<'a, str>>,
-    pub source: Option<Cow<'a, str>>,
-    pub curr_line: u32,
-    pub line_defined: u32,
-    pub last_line_defined: u32,
-    pub num_ups: u32,
-    pub num_params: u32,
-    pub is_vararg: bool,
-    pub is_tailcall: bool,
-    pub short_src: Option<Cow<'a, str>>,
-
-    #[doc(hidden)]
-    _unused: ()
+    ar: *mut lua_Debug,
+    state: *mut lua_State,
+    _phantom: PhantomData<&'a ()>
 }
 
 impl<'a> Debug<'a> {
-    /// Construct a new `Debug` structure that is not associated with a Lua debug structure. It
-    /// involves some string copying.
-    pub fn to_owned(&'a self) -> Debug<'static> {
-        Debug {
-            name: self.name.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
-            namewhat: self.namewhat.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
-            what: self.what.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
-            source: self.source.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
-            curr_line: self.curr_line,
-            line_defined: self.line_defined,
-            last_line_defined: self.last_line_defined,
-            num_ups: self.num_ups,
-            num_params: self.num_params,
-            is_vararg: self.is_vararg,
-            is_tailcall: self.is_tailcall,
-            short_src: self.short_src.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
-            _unused: ()
+    /// Corresponds to the `n` what mask.
+    pub fn names(&self) -> Names<'a> {
+        unsafe {
+            if ffi::lua_getinfo(self.state, cstr!("n"), self.ar) == 0 {
+                rlua_panic!("lua_getinfo failed with `n`")
+            } else {
+                Names {
+                    name: ptr_to_str((*self.ar).name as *const i8),
+                    name_what: ptr_to_str((*self.ar).namewhat as *const i8)
+                }
+            }
         }
     }
+
+    /// Corresponds to the `n` what mask.
+    pub fn source(&self) -> Source<'a> {
+        unsafe {
+            if ffi::lua_getinfo(self.state, cstr!("S"), self.ar) == 0 {
+                rlua_panic!("lua_getinfo failed with `S`")
+            } else {
+                Source {
+                    source: ptr_to_str((*self.ar).source as *const i8),
+                    short_src: ptr_to_str((*self.ar).short_src.as_ptr() as *const i8),
+                    line_defined: (*self.ar).linedefined as i32,
+                    last_line_defined: (*self.ar).lastlinedefined as i32,
+                    what: ptr_to_str((*self.ar).what as *const i8),
+                }
+            }
+        }
+    }
+
+    /// Corresponds to the `l` what mask. Returns the current line.
+    pub fn curr_line(&self) -> i32 {
+        unsafe {
+            if ffi::lua_getinfo(self.state, cstr!("l"), self.ar) == 0 {
+                rlua_panic!("lua_getinfo failed with `l`")
+            } else {
+                (*self.ar).currentline as i32
+            }
+        }
+    }
+
+    /// Corresponds to the `t` what mask. Returns true if the hook is in a function tail call, false
+    /// otherwise.
+    pub fn is_tail_call(&self) -> bool {
+        unsafe {
+            if ffi::lua_getinfo(self.state, cstr!("t"), self.ar) == 0 {
+                rlua_panic!("lua_getinfo failed with `t`")
+            } else {
+                (*self.ar).currentline != 0
+            }
+        }
+    }
+
+    /// Corresponds to the `u` what mask.
+    pub fn stack(&self) -> Stack {
+        unsafe {
+            if ffi::lua_getinfo(self.state, cstr!("u"), self.ar) == 0 {
+                rlua_panic!("lua_getinfo failed with `u`")
+            } else {
+                Stack {
+                    num_ups: (*self.ar).nups as i32,
+                    num_params: (*self.ar).nparams as i32,
+                    is_vararg: (*self.ar).isvararg != 0
+                }
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Names<'a> {
+    pub name: Option<&'a str>,
+    pub name_what: Option<&'a str>
+}
+
+#[derive(Clone, Debug)]
+pub struct Source<'a> {
+    pub source: Option<&'a str>,
+    pub short_src: Option<&'a str>,
+    pub line_defined: i32,
+    pub last_line_defined: i32,
+    pub what: Option<&'a str>
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct Stack {
+    pub num_ups: i32,
+    pub num_params: i32,
+    pub is_vararg: bool
 }
 
 /// Indicate in which circumstances the hook should be called by Lua.
@@ -129,27 +201,9 @@ impl Default for HookTriggers {
 /// This callback is passed to `lua_sethook` and gets called whenever debug information is received.
 pub(crate) unsafe extern "C" fn hook_proc(state: *mut lua_State, ar: *mut lua_Debug) {
     callback_error(state, || {
-        if ffi::lua_getinfo(state, cstr!("nSltu"), ar) == 0 {
-            rlua_panic!("lua_getinfo failed")
-        }
-
         let lua = Lua::make_ephemeral(state);
         let debug = Debug {
-            name: ptr_to_str((*ar).name as *const i8),
-            namewhat: ptr_to_str((*ar).namewhat as *const i8),
-            what: ptr_to_str((*ar).what as *const i8),
-            source: ptr_to_str((*ar).source as *const i8),
-            curr_line: (*ar).currentline as u32,
-            line_defined: (*ar).linedefined as u32,
-            last_line_defined: (*ar).lastlinedefined as u32,
-            num_ups: (*ar).nups as u32,
-            num_params: (*ar).nparams as u32,
-            is_vararg: (*ar).isvararg == 1,
-            is_tailcall: (*ar).istailcall == 1,
-            short_src: str::from_utf8(mem::transmute((*ar).short_src.as_ref()))
-                .and_then(|r| Ok(Some(Cow::from(r))))
-                .unwrap_or(None),
-            _unused: ()
+            ar, state, _phantom: PhantomData
         };
 
         let cb = (&*extra_data(state)).hook_callback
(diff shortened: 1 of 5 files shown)


src/hook.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub source: Option<Cow<'a, str>>,
+    pub curr_line: u32,
+    pub line_defined: u32,
+    pub last_line_defined: u32,
+    pub num_ups: u32,
+    pub num_params: u32,
+    pub is_vararg: bool,
+    pub is_tailcall: bool,
+    pub short_src: Option<Cow<'a, str>>
+}
```
jonas-schievink: Consider adding a private field here to allow rlua to add more fields without breaking users

src/hook.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub calls: bool,
+    /// When Lua returns from a function.
+    pub returns: bool,
+    /// Before executing a new line, or returning from a function call.
+    pub lines: bool,
+    /// After a certain amount of instructions specified by `count`.
+    pub after_counts: bool,
+    /// Specify how many instructions to execute before calling the hook. Only effective when
+    /// `after_counts` is set to true.
+    pub count: u32
```
jonas-schievink: It might make sense to turn `after_counts` into an `Option<u32>` instead of ignoring this field when `after_counts` is `false`

src/lua.rs
```diff
@@ -625,6 +626,69 @@ impl Lua {
...
+    /// let lua = Lua::new();
+    /// lua.set_hook(HookOptions {
+    ///     lines: true, ..Default::default()
+    /// }, |debug: &Debug| {
+    ///     println!("line {}", debug.curr_line);
+    ///     Ok(())
+    /// });
+    /// let _ = lua.exec::<_, ()>(code, None);
+    /// # }
+    /// ```
```
jonas-schievink: Nice docs!

src/lua.rs
```diff
@@ -625,6 +626,69 @@ impl Lua {
...
+    ///
+    /// # Example
+    ///
+    /// Shows each line number of code being executed by the Lua interpreter.
+    ///
+    /// ```
+    /// # extern crate rlua;
+    /// # use rlua::{Lua, HookOptions, Debug};
+    /// # fn main() {
+    /// let code = r#"local x = 2 + 3
```
jonas-schievink: Ultra-nitpick: try putting a newline after the `r#"` and indent the Lua code to make it look a bit nicer

src/lua.rs
```diff
@@ -625,6 +626,69 @@ impl Lua {
...
+        F: 'static + Send + Fn(&Debug) -> Result<()>
+    {
+        self.set_mut_hook(options, callback)
+    }
+
+    /// Set a function for Lua the same way as [`set_hook`], but allows the use of a mutable
+    /// function. The behavior is the same otherwise.
+    ///
+    /// [`set_hook`]: #method.set_hook
+    pub fn set_mut_hook<F>(
```
jonas-schievink: I think this method is redundant, since every closure that implements `Fn` also implements `FnMut`, so just having one `set_hook` that takes an `FnMut` should be enough.

src/lua.rs
lemarcuspoilus: I did not since Lua would report it being on the 2nd line (2-3-4 would be outputted) and that looked confusing to me. Might change it to something else.

lemarcuspoilus:
I applied the changes you've recommended to me with some extras of my own. I have no idea why some of them didn't come to mind as I wrote the code!

While tweaking the hook code, I realized I could write the execution time limit code right within the Lua struct and the `hook_proc` function. What do you think of that? As long as the hook triggers are set to call the hook on a regular basis, this could work.

src/hook.rs
```diff
@@ -0,0 +1,162 @@
...
+            is_tailcall: (*ar).istailcall == 1,
+            short_src: str::from_utf8(mem::transmute((*ar).short_src.as_ref()))
+                .and_then(|r| Ok(Some(Cow::from(r))))
+                .unwrap_or(None),
+            _unused: ()
+        };
+
+        let cb = extra.hook_callback
+            .as_mut()
+            .expect("no hooks previously set; this is a bug");
```
kyren: I think there should probably be an `rlua_expect` macro inside the library that prints consistent messages to match `rlua_assert` and `rlua_abort` and that `rlua_expect` should be used here and also everywhere that currently `.unwrap()` is used.  I also like that you say "this is a bug", and I think that all of the error macros in macros.rs should also say that.

kyren:
Oh wow, this PR is great, thank you so much!

I agree with all the feedback @jonas-schievink gave so thank you for taking the time to review this before I had a chance to 👍

The only feedback I have to add atm is a super nit about using expect that I added some comments for.

kyren:
You're free to either add that `rlua_expect` macro that I mentioned and change the error messages, or since that's more of a style concern you could also just leave that to me to change, either way.

Other than that I don't really see any reason right now why I shouldn't merge this, this is really good work!

lemarcuspoilus:
Hello! If you don't mind, I'd like this PR to stay focused on the hook code; by that I mean I wouldn't add the `rlua_expect!` just yet so I don't start making changes everywhere ;)  I think it would be a better idea to make a `LuaExpect` trait or something like that which adds the `rlua_expect` method to `Result`. It would result in a very similar, clean API like `expect`.

As of the names, I did feel that `after_counts` was slightly odd. I originally wanted them to sound like _trigger on returns_, _trigger on lines_ (hmmm maybe not that one) and so on. Since the hook is called repeatedly, I think `returns` suggests it is for many returns and not just for a specific return somewhere. I'm still not sure about how I could replace `after_counts` with something more explanatory, yet short. What about `every_nth_instructions`?

I appreciate the positive feedback about the quality of the PR <3 at the same time I feel it's an expected, implicit standard.

kyren:
> Hello! If you don't mind, I'd like this PR to stay focused on the hook code; by that I mean I wouldn't add the rlua_expect! just yet so I don't start making changes everywhere ;) I think it would be a better idea to make a LuaExpect trait or something like that which adds the rlua_expect method to Result. It would result in a very similar, clean API like expect.

Yeah that makes complete sense, you don't have to do it, it was mostly that just seeing it made me think about the fact that it needs to be done.

> As of the names, I did feel that after_counts was slightly odd. I originally wanted them to sound like trigger on returns, trigger on lines (hmmm maybe not that one) and so on. Since the hook is called repeatedly, I think returns suggests it is for many returns and not just for a specific return somewhere. I'm still not sure about how I could replace after_counts with something more explanatory, yet short. What about every_nth_instructions? (not sure about the s)

Okay, yeah I can see what you mean about implying that the hook is called repeatedly, `on_returns` meaning "on *every* return, which there will be multiple of".  How about `on_calls` `on_returns`, `every_line`, `every_nth_instruction`?

> I appreciate the positive feedback about the quality of the PR <3 at the same time I feel it's an expected, implicit standard.

Just to make sure I understand, you mean it's an expected implicit standard that the PRs be good haha?  I suppose so, but I'd still like to acknowledge that you met it!

lemarcuspoilus:
Just mixed the plurality currently being used with your suggestion. I think they all match pretty well!
There are a few things on top of my head that could be interesting to add before merging:

- I just remembered there is a `Scope` structure there. Hooks look like a good fit since, as least me, I'd use them only for some specific code in most cases
- As I already suggested, would adding time limitation right inside the crate add too much overhead/weight to the crate? I don't think many users would match both normal debugging and time limitation together, but that would imply adding a more user-friendly `TimedOut` error instead of using `ExternalError` or `RuntimeError`.

kyren:
> every_line looks better than each_line for sure. I'd like to hear from @jonas-schievink about those names before making a final commit.

Yeah, agreed.

lemarcuspoilus:
> I don't like super like adding timeouts on wall clock time or doing anything that gives the user the impression that there's more safety than there really is. There is an obvious implementation of such a function with respect to Lua::set_hook, but the obvious implementation has sort of fiendishly subtle real world behavior. If you set an execution timeout on X seconds there's no guarantee that you will trigger the error within X seconds because Lua is allowed to do so much. I also don't like that calling Lua::set_exec_time_limit would have to clear any existing hook. I think if you let the user write the simple obvious implementation then they're responsible for the subtle behavior that it actually has :)

It's true that the precision would be very low if the time between hooks is large. I agree this would be a bad idea.

> Scoped hooks would allow you to expire the hook in a natural way, and would allow for hooks that aren't 'static, but I think adding it would be a bit weird without support for multiple hook functions, since adding a scoped callback would remove any global one. I think it might be very tricky to support multiple hook functions as well, and I'm not sure how much of a use non-'static hook functions would have? Maybe I'm not imagining the use cases very well!

I'm especially interested for hooks that expires because I think it might look more "elegant" in some way to not use `remove_hook` after only one `Lua::exec` call? Not a very compelling argument.

The rather large problem I see with using multiple hooks is that Lua only allows one hook to be set at one given time and _only one `mask` for it._ Since `Lua::set_hook` forwards the triggers to `lua_sethook`, that means we'd need to find the lowest common denominator triggers for all current hooks, then for every call of the hook, check which functions are supposed to be called. Here's some steps of what I mean:

- Hook A gets called every 10th instruction and on function calls
- Hook B gets called every 5th instruction
- Hook C gets called every line

This means the lowest common denominator would be: every line, every 5h instruction, function calls. Hook C would somehow need to be called when a line change is detected (possible through the `Debug::curr_line` field), Hook B every 5th invocation of the callback, and Hook A every other time Hook B is elligible for the call _and_ on function calls. It's totally possible! But I'm worried about all the overhead this would have, especially if `every_nth_instruction` is set to 1.

But don't get me wrong, I'm willing to try to make this happen if you agree!

kyren:
> It's true that the precision would be very low if the time between hooks is large. I agree this would be a bad idea.

It's worse than that in the presence of 1) other callbacks, 2) lack of memory limits, 3) unsafe lua stdlib functions, etc etc etc...  I'm trying to be *ultra* careful about not giving people a false sense of security.

> I'm especially interested for hooks that expires because I think it might look more "elegant" in some way to not use remove_hook after only one Lua::exec call? Not a very compelling argument.

Yeah, I agree that having hooks expire based on a handle is in some way nicer, I just don't like how it interacts with being limited to a single global hook.

So yeah, you could add support for multiple hooks, but doing so would have surprising and weird or deeply magical behavior so I didn't want to suggest that you do it.  I don't think it's a good place to start, at least, though we may want to look at it later.

> The rather large problem I see with using multiple hooks is that Lua only allows one hook to be set at one given time and only one mask for it. Since Lua::set_hook forwards the triggers to lua_sethook, that means we'd need to find the lowest common denominator triggers for all current hooks, then for every call of the hook, check which functions are supposed to be called. Here's some steps of what I mean:

The immediate thing I thought about for implementing multiple hooks based on a single hook is exactly what you mentioned about having multiple Nth instruction hooks.  What if you have two large Ns which are coprime, do you have to then have a hook on every instruction?  Do you do something magical like limit the instruction granularity to some fixed amount?  I couldn't come up with anything that made sense to me to suggest, so I just didn't suggest that we do it... at least yet.

I think the simplest possible implementation of a single global callback hook is okay for now, because it's really clear what the translation to the Lua C API actually is, and it doesn't imply any safety that's not really there.

src/hook.rs
```diff
@@ -0,0 +1,162 @@
...
+///     on_every_nth_instruction: Some(1), ..Default::default()
+/// };
+/// # let _ = triggers;
+/// # }
+/// ```
+pub struct HookTriggers {
+    /// Before a function call.
+    pub on_calls: bool,
+    /// When Lua returns from a function.
+    pub on_returns: bool,
```
jonas-schievink: It might be a good idea to document that this does *not* include returns from tail-called functions:

> For call events, event can be LUA_HOOKCALL, the normal value, or LUA_HOOKTAILCALL, for a tail call; in this case, there will be no corresponding return event.

jonas-schievink:
One more thing worth testing for might be that the hooks are always dropped when `set_hook` or `remove_hook` is called.

Regarding the hook trigger names, I think I like @kyren's last suggestion most:

> `on_calls`, `on_returns`, `every_line`, `every_nth_instruction`

Although I also wouldn't mind `on_call` and `on_return` instead - the docs should make it immediately clear that this means *every call* and *every return* (except returns from tail-called functions, as I noted above).

I also think that timeouts belong in user code, since it would complicate `rlua`'s internals quite a bit and add lots of assumptions. Execution time also doesn't always seem like a *desirable* thing to limit because it's very unpredictable, so instruction limits might be preferable.

Regarding support for multiple hooks: This also seems like something we could reasonably expect a user to implement themselves if they really want it, as it makes the implementation in rlua pretty complicated. A user would know exactly when the "base hook" needs to be called and can dispatch from there.

kyren:
After thinking about it for a while, I think the current hook code is unsound, but it's hopefully pretty easy to fix.

I think the problem lies in the fact that hooks are FnMut, and by using something like Rc<Lua> I believe I can get mutable aliasing in the hook function.  We could either have just Fn hook types or maybe since it's so incredibly rare to recurse inside a hook, we could guard it with a RefCell.  Even though it's rare, I'd lean towards just using Fn because otherwise technically it should be documented that callback hooks use RefCell and it's yet more strange corner case behavior to have to worry about.

As extra paranoia, there's a `&mut ExtraData` reference in `hook_proc` that should probably just be replaced with something like:

```
        let cb = (*extra_data(self.state)).hook_callback
             .as_mut()
             .expect("rlua internal error: no hooks previously set; this is a bug");
         cb(&debug)
```

I know constantly doing `(*extra_data(self.state))` is dumb, but I'm also a bit paranoid about even technically violating mutable aliasing rules (so keeping &mut references alive for as little time is possible is preferred, or never even creating them).

So, actually that would leave us with *another* soundness issue, which is that a hook can call `Lua::remove_hook` inside itself and remove its own storage while executing, but this is just another way of describing another mutable aliasing violation.  Basically, it's just too dangerous to keep a reference to `ExtraData` around across a callback as it's accessed everywhere, even a reference to just a field of `ExtraData`.

I propose that we make the `hook_callback` field into `Option<Rc<Fn(&Debug)>>`, allowing us to avoid any mutable aliasing violations.

lemarcuspoilus:
Hello, wow lots to read!

> It might be a good idea to document that this does not include returns from tail-called functions:
For call events, event can be LUA_HOOKCALL, the normal value, or LUA_HOOKTAILCALL, for a tail call; in this case, there will be no corresponding return event.

The Lua's documentation looks confusing to me: what "in this case" refers to? The latter? Then we have the hook being calls for returns even on tail calls. At least that's what I understand.

I did not add it since the list of event masks visible in `lua.h:412-415` does not include the tail version and only `LUA_MASKCALL`. I eventually forgot about that detail. I should probably check if the behaviour would change.

> One more thing worth testing for might be that the hooks are always dropped when set_hook or remove_hook is called.

I shall add a test for a second call of `set_hook`, but `remove_hook` is covered.

> Regarding the hook trigger names, I think I like @kyren's last suggestion most:

Noted, that's what I'll put for the names. Maybe clarify the doc too, especially about overheads.

> Although I also wouldn't mind on_call and on_return instead - the docs should make it immediately clear that this means every call and every return (except returns from tail-called functions, as I noted above).
> 
> I also think that timeouts belong in user code, since it would complicate rlua's internals quite a bit and add lots of assumptions. Execution time also doesn't always seem like a desirable thing to limit because it's very unpredictable, so instruction limits might be preferable.
> 
> Regarding support for multiple hooks: This also seems like something we could reasonably expect a user to implement themselves if they really want it, as it makes the implementation in rlua pretty complicated. A user would know exactly when the "base hook" needs to be called and can dispatch from there.

It's very cool that the user is not dependent on rlua's decision about it, so I wouldn't mind if we keep it the way it is now, which is nothing more than a fancy wrapper. Instruction limit is also a possibility thought depending on `every_nth_instruction`, the precision would vary.

I am still not against for such a system, but maybe inside some kind of module or even a separate crate. We/I could use a trait to add extra methods for that. You know, I like convenience. :) That being said I'm cool going your route.

> So, actually that would leave us with another soundness issue, which is that a hook can call Lua::remove_hook inside itself and remove its own storage while executing, but this is just another way of describing another mutable aliasing violation. Basically, it's just too dangerous to keep a reference to ExtraData around across a callback as it's accessed everywhere, even a reference to just a field of ExtraData.
>
> I propose that we make the hook_callback field into Option<Rc<Fn(&Debug)>>, allowing us to avoid any mutable aliasing violations, leaving us with something like

You know, I might be a foolish person but I'm chill working with raw pointers, `mem::transmute` and such, I semi-don't understand your worries about having a `&mut` reference. :D

What I do share is the safety concerns of `remove_hook` within a hook. I saw that too but I did not realise it would deallocate the memory being used by the function. I'm going to wrap it inside a `Rc`. As for usign  `Fn` over `FnMut`, I'd prefer to keep the mutability advantage that comes with `FnMut` but that would require the `RefCell` you've mentioned. As for the rest, hmmm I did not catch it quite but I trust you.

lemarcuspoilus:
Just tried to call `remove_hook` within a hook but Rust outputs this error message:
```
error[E0277]: `*mut rlua::ffi::lua_State` cannot be shared between threads safely
   --> tests\hooks.rs:120:9
    |
120 |     lua.set_hook(HookTriggers {
    |         ^^^^^^^^ `*mut rlua::ffi::lua_State` cannot be shared between threads safely
    |
    = help: within `rlua::Lua`, the trait `std::marker::Sync` is not implemented for `*mut rlua::ffi::lua_State`
    = note: required because it appears within the type `rlua::Lua`
    = note: required because of the requirements on the impl of `std::marker::Send` for `&rlua::Lua`
    = note: required because it appears within the type `[closure@tests\hooks.rs:122:8: 126:6 lua:&rlua::Lua, call_count:&mut i32]`
```

I'm using this code:
```rust
    lua.set_hook(HookTriggers {
        every_line: true, ..Default::default()
    }, |_debug: &Debug| {
        lua.remove_hook();
        call_count += 1;
        Ok(())
    });
```

I might have made a mistake but that technically acts as a safety protection without the need to use `Rc<RefCell<...>>`.

kyren:
> You know, I might be a foolish person but I'm chill working with raw pointers, mem::transmute and such, I semi-don't understand your worries about having a &mut reference. :D

It is not even allowed, afaict, for multiple &mut references to the same place toeven *exist at the same time* even if you don't access them.  I don't know for sure because the rules are still being written, but no need to push it.  Lua is insane and requires everything be always re-entrant constantly, so even creating a &mut reference to something "global" has to be done very very carefully, and that's what's effectively causing the unsoundness right now.

> What I do share is the safety concerns of remove_hook within a hook. I saw that too but I did not realise it would deallocate the memory being used by the function. I'm going to wrap it inside a Rc. As for usign Fn over FnMut, I'd prefer to keep the mutability advantage that comes with FnMut but that would require the RefCell you've mentioned. As for the rest, hmmm I did not catch it quite but I trust you. I appreciate the care you both put in the crate to ensure it's basically 100% stable for even the noobest user (which might be me!).

Or set_hook, or calling into more lua which recursively calls the hook.  You can't call a FnMut function a second time while the first call is not finished because the FnMut could have an internally mutable variable and that's mutable aliasing.

> I might have made a mistake but that technically acts as a safety protection without the need to use Rc<RefCell<...>>. Did I missed something?

What is holding the `lua`?  Is it an Arc?  Can I see the whole example?

kyren:
> As for the rest, hmmm I did not catch it quite but I trust you. I appreciate the care you both put in the crate to ensure it's basically 100% stable for even the noobest user (which might be me!).

It should be 100% safe for everyone, it should be 100% safe even in the presence of somebody *trying* to cause UB with safe code, it should be completely 100% safe without the use of unsafe because *that's the contract*.

It's really, really, really hard to make C or C++ or unsafe Rust safe, and we can't and shouldn't rely on any behavior that takes place in safe code.  Safe code is our safe space, it's where we don't have to worry about UB no matter what we do, even if it's weird.  If it compiles, it's safe.

But don't feel bad about not understanding how something is subtly unsafe, it's just really hard to catch :(  I was already beating myself up already for not seeing this yesterday when I first looked at it, but even if I did see this immediately, I've already written a whole bunch of unsoundness into this crate before.  Just go take a look at all the fixed unsoundness issues haha.

I'm pretty confident this is *the* safest Lua bindings layer I've ever used, but I still am really really not sure that it's actually safe.  I would never, even with perfect script sandboxing and memory limits, let this bindings system in its current state interpret Lua code that was controllable by a real attacker.  I'm not even sure I would *ever* trust it to do that, though that should be the eventual goal.

lemarcuspoilus:
Hello, let me go through this!

> It is not even allowed, afaict, for multiple &mut references to the same place to even exist at the same time even if you don't access them. I don't know for sure because the rules are still being written, but no need to push it. Lua is insane and requires everything be always re-entrant constantly, so even creating a &mut reference to something "global" has to be done very very carefully, and that's what's effectively causing the unsoundness right now.

Indeed Rust is very strict on how you access memory. I can see your point and that would explain further why every function in `Lua` is immutable. I wouldn't consider Lua as made for security in any way with all the stack magic nonsense with barely any restriction or structure in how you do things (I mean just making a table is like using the assembly language).

> Or set_hook, or calling into more lua which recursively calls the hook. You can't call a FnMut function a second time while the first call is not finished because the FnMut could have an internally mutable variable and that's mutable aliasing.

I probably overlooked your previous comment, but this makes much more sense to me. I didn't thought about executing more Lua code within a Lua hook. That's enough to convince me to use `Fn` and as immutable code for the hook as possible.

> What is holding the lua? Is it an Arc? Can I see the whole example?
> 
> Edit: oh wait, okay I know why that error happens, you're just capturing the lua by reference I think. Put the lua inside an Arc, then clone the Arc and move it into the hook closure to see what I mean.

Can't believe I did not think about using an `Arc`... hmmm I can't try right now but I'm still not sure it'll satisfy the `Sync` requirement the compiler is screaming at me. I think a `RwLock` would be a decent fix.

> It should be 100% safe for everyone, it should be 100% safe even in the presence of somebody trying to cause UB with safe code, it should be completely 100% safe without the use of unsafe because that's the contract. [...]

Safety is indeed one of Rust's primary goal and shown in your face as you go download rustup, but at the same time I believe it's just not possible to ensure 100% safety. I am in favor of making the code as safe and secure as possible (with as less performance penalty as possible too), don't get me wrong, but you have so many variables going on there (Rust itself, your OS, Lua's C code to name a few) that no guarantees can be made.

I'm so happy that Rust and excellent crates exist so I can actually program without being annoyed with difficult-to-track segmentation faults. And I'm happy that I'm not alone trying to figure out if my code is bulletproof or not! I'm the kind of guy who `memset` a virtual C++ class to zero and doesn't know why the code crashes.

I'll fix this comment and apply your suggestions to the code later on.

lemarcuspoilus:
I am not sure how you are supposed to access Lua from the hook since my tests shows that `RwLock` doesn't add the Sync trait required by the raw pointer.

The only safe-Rust way I've found is to use `Arc<Mutex<Lua>>`, as seen in this test I just made:
```rust
    let code = r#"
        local x = 1
        x = 2
    "#;
    let lua = Arc::new(Mutex::new(Lua::new()));

    let hook_lua = Arc::clone(&lua);
    lua.lock().unwrap().set_hook(HookTriggers {
        every_line: true, ..Default::default()
    }, move |_debug| {
        hook_lua.try_lock().unwrap().remove_hook();
        Err(Error::RuntimeError("the error will show the presence of a hook".to_string()))
    });

    assert!(lua.try_lock().unwrap().exec::<_, ()>(code, None).is_err());
    {
        let lua = lua.try_lock().unwrap();
        assert!(lua.exec::<_, ()>(code, None).is_ok());
        assert_eq!(lua.globals().get::<_, i64>("x").unwrap_or(-1), 2);
    }
```

It does compile, which is pretty cool, but on runtime we have a lock error of `WouldBlock` caused by this line:
```rust
hook_lua.try_lock().unwrap().remove_hook();
```

This happens because `lua` is already being accessed by `lua.try_lock().unwrap().exec::<_, ()>(code, None)`. Unless you succeed in achieving a working solution for this test, I feel assure it's not possible to access the same Lua instance from inside at all (at best the change can be differed with some kind of channel between the hook and the outside).

kyren:
Yeah, sorry for giving you that suggest, you can't put Lua in an Arc because it would require Lua to be Sync in order for the `Arc<Lua>` to be Send, that won't work.

What will work though is TLS, hold on and I'll give you an example.

kyren:
```rust
#[test]
fn bad_hook() {
    thread_local! {
        pub static GLOBAL_LUA: Lua = Lua::new();
    }

    let mut hook_var = 0;
    GLOBAL_LUA.with(|lua| {
        lua.globals().set("i", 1).unwrap();

        lua.set_hook(
            HookTriggers {
                on_calls: true,
                on_returns: true,
                every_line: true,
                every_nth_instruction: Some(1),
                ..Default::default()
            },
            move |_: &Debug| {
                println!(
                    "hook start, address of hook_var is: {:?}",
                    &mut hook_var as *mut i32
                );
                GLOBAL_LUA.with(|lua| {
                    lua.exec::<_, ()>(
                        r#"
                            while true do
                                i = i + 1
                            end
                        "#,
                        None,
                    ).unwrap()
                });
                println!("hook end");
                Ok(())
            },
        );

        lua.exec::<_, ()>("while true do end", None).unwrap();
    });
}
```

Okay, so it looks like Lua 5.3 has some undocumented behavior which actually *makes sense*, but it is a bit unlike how Lua usually operates :D

If you run this example, you'll find that if you execute Lua from within a hook that it seems to work correctly, but it won't trigger further recursive hooks.  Usually Lua loves calling your C code recursively even when you don't expect it haha.

It ***MAY*** be that the semantics of ffi::lua_sethook protect us from the danger of having recursive FnMut calls, but holy hell does that make me nervous and scared.  I can't manage to make it happen here myself, but this behavior is afaict not even documented in the Lua 5.3 manual, so I don't know whether me not being able to make it happen means it's impossible.

This is independent of the fact that you can still set the hook again and that's 100% unsound and needs to be fixed, but I'm considering that part the "easy" part.

kyren:
> Safety is indeed one of Rust's primary goal and shown in your face as you go download rustup, but at the same time I believe it's just not possible to ensure 100% safety. I am in favor of making the code as safe and secure as possible (with as less performance penalty as possible too), don't get me wrong, but you have so many variables going on there (Rust itself, your OS, Lua's C code to name a few) that no guarantees can be made.

Yeah you're right that esp something like rlua might not be able to "guarante" safety because it would have to also guarantee the correctness of PUC-Rio Lua 5.3, but that doesn't mean that I'll abandon the weaker but still very useful guarantee that:

If you find a way to cause UB with safe code I will go to as far as I humanly can to fix it, even if that safe code is weird or bad or malicious or ridiculous, because striving to maintain the safety contract is just that important.  Even if you *can't* guarantee for whatever reason that there is no memory unsafety because of reliance on flawed operating systems, underlying libraries etc, it's *still useful* because if you see memory unsafety *you know the bug is not in your safe code*.

But yeah, if you were to need a Lua interpreter you could more easily verify as genuinely safe, you might have to wait for `luster`.

lemarcuspoilus:
Hmmmm I thiiiink I understand what is `threadlocal!` and that looks pretty hacky. Let me ask you, will it consider the hook as another thread? I just ran your code (modified to this):

```rust
thread_local! {
        pub static GLOBAL_LUA: Lua = Lua::new();
    }

    let mut hook_var = 0;
    GLOBAL_LUA.with(|lua| {
        lua.globals().set("i", 1).unwrap();

        lua.set_hook(
            HookTriggers {
                on_calls: true, on_returns: true, every_line: true, every_nth_instruction: Some(1),
                ..Default::default()
            },
            move |_: &Debug| {
                println!(
                    "hook start, address of hook_var is: {:?}",
                    &mut hook_var as *mut i32
                );
                GLOBAL_LUA.with(|lua| {
                    lua.exec::<_, ()>(
                        r#"
                            for j = 1, 10 do
                                i = i + 1
                            end
                        "#,
                        None,
                    ).unwrap()
                });
                println!("hook end");
                Ok(())
            },
        );

        println!("BEGINNING OF TIME");
        lua.exec::<_, ()>("for j = 1, 10 do end", None).unwrap();
        println!("{}", lua.globals().get::<_, i64>("i").unwrap());
    });
```

... I get the following:
```
BEGINNING OF TIME
hook start, address of hook_var is: 0x4a9b438d28
hook end
hook start, address of hook_var is: 0x4a9b438d28
hook end
hook start, address of hook_var is: 0x4a9b438d28
hook end
[...]
hook start, address of hook_var is: 0x4a9b438d28
hook end
hook start, address of hook_var is: 0x4a9b438d28
hook end
371
```

I have a good idea for a fix: 
```rust
pub(crate) struct ExtraData {
    registered_userdata: HashMap<TypeId, c_int>,
    registry_unref_list: Arc<Mutex<Option<Vec<c_int>>>>,

    ref_thread: *mut ffi::lua_State,
    ref_stack_size: c_int,
    ref_stack_max: c_int,
    ref_free: Vec<c_int>,

    pub hook_callback: Option<Rc<RefCell<FnMut(&Debug) -> Result<()>>>>
}
```
What we can do is add a boolean that will be set to true whenever code is being ran (either by `load`, `exec` or `eval`, anything) so that it would be impossible to evaluate more code within a hook. That being said we prevent the user from using such a feature. A `Rc` would remove any risk of `set_hook` or `remove_hook` to break the system. At worse, it will only switch the hook after the first one is done being executed.

I'll try to implement all of this so we can experiment on top of new code.

kyren:
> Hmmmm I thiiiink I understand what is threadlocal! and that looks pretty hacky. Let me ask you, will it consider the hook as another thread? I just ran your code (modified to this):

It's not a thread, it's an interface to TLS (Thread Local Storage).  It's a "global", except one independent instance exists per thread, so there's no need to require it be `Sync`.  It is not hacky, it is a normal use for TLS and I've seen projects use a global `rlua` inside TLS.

> ... I get the following:

What you're seeing is that the inner hooks are *not* being called, you can change the example to print things inside the inner and outer loops to see when hooks are called.  It appears that triggering a hook while inside another hook simply does not execute the hook, which currently saves us from unsoundness w.r.t. `FnMut.  I'm unsure of what happens if you call lua_sethook while already inside a hook, possibly it works as expected but allows you to recurse, possibly it is UB, I don't know.

> I have a good idea for a fix:

Like I said, I don't even know that it's actually broken, but I don't think we should rely on the apparent current behavior of hooks inside other hooks, as it's unspecified in the manual.

(Again, I mean the unsoundness due to recursive FnMuts, the unsoundness due to mutating the callback while it's being called still exists and is separate)

> What we can do is add a boolean that will be set to true whenever code is being ran (either by load, exec or eval, anything) so that it would be impossible to evaluate more code within a hook. That being said we prevent the user from using such a feature. A Rc would remove any risk of set_hook or remove_hook to break the system. At worse, it will only switch the hook after the first one is done being executed.

I'm not following why a boolean would be necessary, we can just use `Option<Rc<Fn(&Debug) -> Result<()>>>` and everything will be fine.  That would solve *both* sources of unsoundness at once, is there something that you particularly don't like about that solution?  We don't need to prevent users from evaluating more code within a hook, that limitation would be strange and hard to describe and verify, why would we do that when we can just ensure that it is safe?

> By the way, it's true that recoding the whole Lua interpreter would give us much more control of what's happening, but that's just a bad idea. Lua already has LuaJIT that some projects depend, it doesn't need more ecosystem fragmentation... unless we convince all of them to work on a single implementation. I don't see that happening, at least for a Lua Lua.

Come on, don't just call my project "a bad idea", that's pretty rude, you don't even know why I'm doing it!  There are *tons* of alternate Lua interpreters by the way, usually when bindings systems from C to the host language are painful enough that it's more or less required to reimplement them.  I know of a couple in both Go and Java that exist simply to solve the otherwise unsolvable problems of garbage collectors interacting.

In any case, forget that project exists, its purpose is only to implement Lua specifically, the other two parts are to prove that you can build a real scripting language runtime in (nearly all) safe code that is competitive in performance to C, and to prove that language binding layers do not have to be so intrinsically slow.  We'll see if I'm successful.

FWIW the goal is also really really complete compatibility with PUC-Rio Lua, in contrast with the situation with LuaJIT, and to track new PUC-Rio Lua versions as they come out.

lemarcuspoilus:
Okay, I just pushed a very very experimental code to fix that odd Lua behavior by simply blocking it. _It doesn't work_ for the `hook_removal` test, for some very odd reason. Even atomics didn't fix it. I'll give it a deeper look soon. Feel free to give it a look if you want!

kyren:
> Okay, I just pushed a very very experimental code to fix that odd Lua behavior by simply blocking it. It doesn't work for the hook_removal test, for some very odd reason. Even atomics didn't fix it. I'll give it a deeper look soon. Feel free to give it a look if you want!

I don't think I'm following at all, because I don't know what weird behavior you're talking about.  I assume you haven't read my response above so I'll give you a sec to read that first.

kyren:
> Even atomics didn't fix it.

There aren't any threads involved, I don't know how atomics would do anything?

kyren:
Okay, after reading the Lua 5.3 source, I now believe the behavior is intentional just not mentioned in the manual.  TBH it makes a lot of sense, you might want to implement Lua hooks in terms of Lua itself, so it wouldn't be helpful if hooks could recursively call themselves.

ldo.c:266
`L->allowhook = 0;  /* cannot call hooks inside a hook */`

In that case, I think it would actually be fine to place a FnMut hook callback into a RefCell, and try_write the RefCell and call something like `.expect("Lua should not allow hook callbacks within a hook")`.

I fully understand the desire to use `FnMut` rather than `Fn`, so that seems to solve things nicely without any unexpected limitations we would have to document.

kyren:
Actually, now that I know this behavior is intentional, we should probably actually have hooks take `&Lua` as the other callback functions do, as it's apparently expected and normal for hook functions to recurse into other Lua?

You know that's actually pretty useful and I had no idea you could even do that until just now, huh.

I know this is turning into a more lengthy PR than you may have anticipated, so if you get things working to the point that the hook is safe I can merge and then make that additional change later, but you're also welcome to do it if you'd like.

lemarcuspoilus:
Hello, I don't mind if it's a long PR, I like working with you :) Probably the other way around too.

Right before I jump in to answer your questions, let me apologize for how you perceived my opinion on making a Lua interpreter. I did not think you were actually serious about making such a project happen. Is it what you meant by `luster`? I also missed Java and C# implementations while typing that so I was wrong. Still, it's no small task... are you actually planning to replace `rlua` with a full Lua interpreter or I'm missing something?

> I think we may have been talking past each other haha.

I think so, plus indeed I posted my small message right when you posted yours. I rushed it since time was becoming an issue.

> That actually makes sense, but that's the genuine behavior of PUC-Rio Lua, so I think we should just document what it does rather than trying too hard to change what actually happens, rlua is meant to be "just" a fancy wrapper not to try to be too clever.

That would be a good thing to do. I'll mention it in the `set_hook` function doc.

> It's not a thread, it's an interface to TLS (Thread Local Storage). It's a "global", except one independent instance exists per thread, so there's no need to require it be Sync. It is not hacky, it is a normal use for TLS and I've seen projects use a global rlua::Lua inside TLS.

I've never used it or maybe even seen it often. The doc says so but I wasn't sure what the usage was... that would help me fix the `hook_swap_within_hook` test.

Right before I get too confused, what I wanted to do is exactly what Lua anticipated: prevent recursive calls of the hook just like the example you've shown. And the code is disgusting, not to say the least. As for atomics, I only wanted to make sure my boolean was actually being written to, which it was.

As I can see you look like comfortable simply using `Rc` and trusting Lua to ensure the hook is always being called safely. All that nastiness code I made is out of the window. Now I'll try to unmix it all:

> I fully understand the desire to use FnMut rather than Fn, so that seems to solve things nicely without any unexpected limitations we would have to document.

I don't remember why I wanted `FnMut` so bad. :D I think it's because I wanted to modify variables outside the hook, like `mpsc::channels` and such, or most certainly modify values passed to the function. Rust workings are still not 100% clear to me, but I think it's that.

> In that case, I think it would actually be fine to place a FnMut hook callback into a RefCell, and try_write the RefCell and call something like .expect("Lua should not allow hooks to be called within another hook").

I'm the one jumping around. Anyways, that would be a solution. But then, this panic would mean it's an actual bug. If that's the case, wouldn't an error, then a simple return be the solution? That would mostly be if the user links rlua with their system Lua. I'll do as you suggest for now so we are on the same page.

> Actually, now that I know this behavior is intentional, we should probably actually have hooks take &Lua as the other callback functions do, as it's apparently expected and normal for hook functions to recurse into other Lua?

I'd be tempted to execute more Lua within hooks. I don't know what but I feel it would be very useful! But, how would we get a `Lua` object then? I'm not familiar enough with the crate, but what I'm thinking would be dirty... like allowing creating `Lua` objects unsafely from a simple `lua_State` pointer. I'll try that.

> TBH it makes a lot of sense, you might want to implement Lua hooks in terms of Lua itself, so it wouldn't be helpful if hooks could recursively trigger themselves.

That's a feature provided by the `debug` module. It's legitimately a feature! But it would override the hook from Rust then, the other way around too... I gotta test that too.

lemarcuspoilus:
I'm glad to report it works really well on my side but Travis keeps complaining for some reason.

tests/hooks.rs
```diff
@@ -121,7 +121,7 @@ fn hook_swap_within_hook() {
     lua.set_hook(HookTriggers {
         every_line: true, ..Default::default()
     }, move |lua, _debug| {
-        lua.globals().set("ok", 1i64);
+        let _ = lua.globals().set("ok", 1i64);
```
jonas-schievink: You can just unwrap this

kyren:
> Right before I jump in to answer your questions, let me apologize for how you perceived my opinion on making a Lua interpreter. I did not think you were actually serious about making such a project happen. Is it what you meant by luster? I also missed Java and C# implementations while typing that so I was wrong. Still, it's no small task... are you actually planning to replace rlua with a full Lua interpreter or I'm missing something?

I assumed you knew what `luster` was since you responded without asking, I thought you were talking about it specifically.  It's fine though, no worries, I probably took it as a bit more harsh than you intended, especially if you weren't aware that I was working on such a thing.

> I think so, plus indeed I posted my small message right when you posted yours. I rushed it since time was becoming an issue.

Time isn't an issue, it's fine!

> I'm the one jumping around. Anyways, that would be a solution. But then, this panic would mean it's an actual bug. If that's the case, wouldn't an error, then a simple return be the solution? That would mostly be if the user links rlua with their system Lua. I'll do as you suggest for now so we are on the same page.

If you have a feature that is based on an underlying assumption and then that assumption turns out not to be true, a panic is the right answer.  Err is for Lua script errors and user errors.

In any case it's not a big deal because it's obvious to me now that allowing Lua to be called from hooks is intentional and supported, and that further hooks are disabled during the top level hook.  Like you said, it's even supported via the Lua debug library!  I don't know how else it would have been implemented, I just had never looked at it.

> I'd be tempted to execute more Lua within hooks. I don't know what but I feel it would be very useful! But, how would we get a Lua object then? I'm not familiar enough with the crate, but what I'm thinking would be dirty... like allowing creating Lua objects unsafely from a simple lua_State pointer. I'll try that.

That's already how it's done in other callbacks, that's how it's designed to work.

> This is not really related to the PR but since we are already talking about other stuff, may I know what pushes you to make a brand-new Rust Lua intepreter/JIT? Assuming I'm still not mixing myself since I haven't seen the program you talked about online. I've made one or two language intepreters only to discover how hard it is.

I didn't mean to get into a whole discussion about `luster`, it's way too early to talk about really.  I just mentioned it off hand when talking about safety because the eventual goal there is to be an example of a memory safe language runtime.

The reasons it exists is actually mostly because I was pretty dissatisfied with the Lua C API and wanted something better.  The existing C API has some not insignificant unavoidable overhead, and I find that this actually limits the usefulness of Lua *in general*, and the situation is even worse with a safe bindings system built on top.  When you embed a scripting language you know you're buying into a significant speed penalty for scripted code, but the problem is not the intrinsic speed penalty as much as that with a high overhead bindings system, speed problems in scripts are often *unfixable*.  With a "zero cost" bindings layer, any script code you replace with Rust code is more or less is a win, but with a high overhead bindings layer often times your attempts at performance fixes will just get lost in the FFI overhead.  There's an entire separate C FFI system in LuaJIT just to get around the intrinsic overhead of the Lua C API.

There's a reason that the Lua C API is designed like it is though, and that is to prevent the user from having to deal with garbage collected pointers and maintain garbage collector invariants.  To make a safe zero cost replacement for the Lua C API, you need a safe, zero cost interface to garbage collected pointers, which is a notoriously difficult problem in Rust.  This problem is so fundamental that this is actually the problem that I *started* with, rather than starting by implementing a language runtime.

It turns out that you CAN in fact make a safe garbage collection system in Rust if you accept a pretty significant limitation on garbage collected pointers and the Rust stack.  That's the real reason `luster` exists, to prove to myself that you can make an entire, useful runtime even with this significant limitation, and that you can design a zero cost FFI for such a runtime based around Rust's safety guarantees.  It's too early to get excited or anything though, it's a Lua interpreter that barely even interprets a tiny subset of Lua, mostly right now it's just a complicated garbage collection demo.

For the record I had no plans yet for adding a JIT compiler or anything like that... right now I'm just experimenting to see where this goes.

tests/hooks.rs
kyren: it's a test, if for some reason `lua.globals().set` failed inside a test, we should know about it!

kyren:
> any ideas? I might have a tiny suggestion for memory limits to come next

Memory limits are a bit tricky in the presence of `gc_guard` and its assumptions.  It's not that hard to fix, but even if you made memory limits work but ignore allocations within `gc_guard`, that would leave `gc_guard` as a way to violate memory limits which could be an issue.  I'm not sure how much of an issue it is and how exploitable it would be, but it's a pretty bad footgun.  Securing Lua scripts is already an ocean of footguns though, so maybe I'm worrying too much?

Getting rid of existing uses of `gc_guard` is an API compatibility break, and also *really* annoying, but potentially that's more honest in the presence of memory limits.

kyren:
> Tho, this is a very silly question but, wouldn't the answer be a custom Arc allocating from a heap? I might be oversimplyfing things but this gives me ideas to try working on this (not this on this project it's your playground but somewhere separate).

This is way off topic of course, but Arc:
1) already allocates from the heap
2) cannot collect cycles (unlike the Lua gc and the vast majority of other gcs which can)
3) is not Copy and has a lot of overhead vs a machine pointer, so it's much slower than what language runtimes normally do.

If you're interested in Rust garbage collector design, I recommend reading [this](https://manishearth.github.io/blog/2015/09/01/designing-a-gc-in-rust/) excellent blog post, and then if you're still interested after that and want to know what I was talking about re: cheap Gc pointers with limitations, you can the comments in [this file](https://github.com/kyren/luster/blob/master/gc-arena/src/arena.rs).

lemarcuspoilus:
Thanks for the links, I'll review them later! Yes by _a_ heap I mean like a large pre-allocated memory pool so it cuts a bit of overhead due to a `malloc`.

> Memory limits are a bit tricky in the presence of gc_guard and its assumptions. It's not that hard to fix, but even if you made memory limits work but ignore allocations within gc_guard, that would leave gc_guard as a way to violate memory limits which could be an issue. I'm not sure how much of an issue it is and how exploitable it would be, but it's a pretty bad footgun. Securing Lua scripts is already an ocean of footguns though, so maybe I'm worrying too much?
> 
> Getting rid of existing uses of gc_guard is an API compatibility break, and also really annoying, but potentially that's more honest in the presence of memory limits. (Edit: I'm not sure whether it's actually possible to remove gc_guard soundly though, it might not be possible for all current uses, but all current uses might not be problematic?)

It's actually stupid simple, maybe not great, but you'll see when I make an issue about that.

src/hook.rs
```diff
@@ -0,0 +1,175 @@
...
+            curr_line: (*ar).currentline as u32,
+            line_defined: (*ar).linedefined as u32,
+            last_line_defined: (*ar).lastlinedefined as u32,
+            num_ups: (*ar).nups as u32,
+            num_params: (*ar).nparams as u32,
+            is_vararg: (*ar).isvararg == 1,
+            is_tailcall: (*ar).istailcall == 1,
+            short_src: str::from_utf8(mem::transmute((*ar).short_src.as_ref()))
+                .and_then(|r| Ok(Some(Cow::from(r))))
+                .unwrap_or(None),
```
kyren: Please don't use a transmute here, please do `(*ar).short_src as *const u8` instead.

src/hook.rs
```diff
@@ -0,0 +1,175 @@
...
+        };
+
+        let cb = (&*extra_data(state)).hook_callback
+            .as_ref()
+            .expect("rlua internal error: no hooks previously set; this is a bug");
+        (&mut *match cb.try_borrow_mut() {
+            Ok(b) => b,
+            Err(_) => rlua_panic!("Lua should not allow hooks to be called within another hook;\
+                please make an issue")
+        })(lua, &debug)
```
kyren: This is still UB, because you're not cloning the Rc you're borrowing it.

src/lua.rs
```diff
@@ -71,6 +73,16 @@ impl Lua {
...
 
+    /// Creates a new Lua state from a state already declared. It is assumed `state` is valid.
+    pub(crate) unsafe fn from_state(state: *mut ffi::lua_State) -> Lua {
+        Lua {
+            state,
+            main_state: main_state(state),
+            ephemeral: true,
+            _phantom: PhantomData
+        }
+    }
```
kyren: It's reasonable to want to make this a separate function, but creating an ephemeral Lua is only done in (now) two places. Either both should use this auxiliary function or neither should. Also it should be called something like make_ephemeral.

src/hook.rs
```diff
@@ -0,0 +1,175 @@
...
+unsafe fn ptr_to_str<'a>(input: *const i8) -> Option<Cow<'a, str>> {
+    if input == ptr::null() || ptr::read(input) == 0 {
+        return None;
+    }
+    let len = libc::strlen(input) as usize;
+    let input = slice::from_raw_parts(input as *const u8, len);
+    str::from_utf8(input)
+        .and_then(|r| Ok(Some(Cow::from(r.trim_right()))))
+        .unwrap_or(None)
+}
```
kyren: I don't like that a method called `ptr_to_str` is doing trimming, it shouldn't be modifying the strings that Lua returns.  Also, I didn't think about the fact that `str::from_utf8` is fallible here, so now I'm not sure I like that the Debug structure returns `str` rather than `[u8]`.  I was recently reminded that we shouldn't make the assumption that Lua strings are utf8 and had gotten into the habit of doing that in too many places.

src/lua.rs
```diff
@@ -625,6 +637,59 @@ impl Lua {
...
+
+    /// Remove any hook previously set by `set_hook`. This function has no effect if no hook was
+    /// set.
+    pub fn remove_hook(&self) {
+        unsafe {
+            (*extra_data(self.state)).hook_callback = None;
+            ffi::lua_sethook(self.state, None, 0, 0);
+        }
+    }
```
kyren: The hook callback function should take a `&Lua` otherwise this is unsound.

src/hook.rs
lemarcuspoilus: I think it was because it was a fixed list, not sure if the cast was allowed

src/hook.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl<'a> Debug<'a> {
+    /// Construct a new `Debug` structure that is not associated with a Lua debug structure. It
+    /// involves some string copying.
+    pub fn to_owned(&'a self) -> Debug<'static> {
+        Debug {
+            name: self.name.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
+            namewhat: self.namewhat.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
+            what: self.what.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
+            source: self.source.as_ref().and_then(|s| Some(Cow::Owned(s.as_ref().to_string()))),
```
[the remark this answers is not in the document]
lemarcuspoilus: `.map()` exists? Didn't know it was possible to do that

src/hook.rs
lemarcuspoilus: Forgot about that!

src/lua.rs
lemarcuspoilus: The problem is that the module `hook` cannot access the fields of `Lua` and I didn't want to have `pub(crate)` everywhere. I will rename the function.

src/lua.rs
lemarcuspoilus: Can't remember if I did the change. Might have forgotten to push. I passed it by value since I thought it would not be used anyway in `hook_proc`.

src/hook.rs
kyren: `(*ar).short_src` has type `*const i8`, right?  That's why you're transmuting?

kyren:
I added some more comments for things I think still need fixing.

I think I jumped the gun a bit at the beginning when I first looked at the PR, I need to learn to slow down and fully carefully review as the very first thing I do, rather than trying to skim, especially when it comes to unsafe code.

Like everything around Lua, there was more subtlety around hooks than I had appreciated at first.

lemarcuspoilus:
> I don't like that a method called ptr_to_str is doing trimming, it shouldn't be modifying the strings that Lua returns. Also, I didn't think about the fact that str::from_utf8 is fallible here, so now I'm not sure I like that the Debug structure returns str rather than [u8]. I was recently reminded that we shouldn't make the assumption that Lua strings are utf8 and had gotten into the habit of doing that in too many places.

It's more complicated than that, here's why:
- Depending on where the hook occurs, some strings are set to full zeros because they don't need to have the information.
- It's failable mostly because `str` requires it to, I did not take the time to look for a ASCII version. I preferred to avoid allocations
- Trimming is done because in some cases, the string contained `\u00` and similar characters at the right end. This just removes them so the string is clean
- `&str` has the advantage it's convenient to use rather than an extra `[u8]` conversion, at least if you're comparing byte strings.

I'll make the changes soon when I got time.

src/hook.rs
kyren: Ohhhh, no it's an array, hold on let me think about this some more, I still don't like `mem::transmute`, *especially* without the types being specified

kyren: Okay, I think it would be cleaner to do slice::from_raw_parts like is done in `ptr_to_str` again, and possibly use the same function that the rest of the conversion does.  As a general rule, don't call `mem::transmute` without type annotation, even if you know the types at the time.  It's incredibly dangerous in the presence of refactoring, because if you refactor and change the types `mem::transmute` will happily produce any type you desire that is the same size.

At the very least it should specify the types with turbofish.

kyren:
> Trimming is done because in some cases, the string contained \u00 and similar characters at the right end. This just removes them so the string is clean

Trim removes *all whitespace* not just zeroes, there are other ways to get `str` out of `[u8]` that stops at the first zero that don't involve trimming whitespace.

> It's failable mostly because str requires it to, I did not take the time to look for a ASCII version. I preferred to avoid allocations

I'm not suggesting allocation, I'm suggesting `[u8]` with a set of convenient methods to go to `str`, like `LuaString` has.  I'm not sure exactly what it should look like yet.

> &str has the advantage it's convenient to use rather than an extra [u8] conversion, at least if you're comparing byte strings.

I know str is convenient, but Lua source code is not utf8.  A user should at least have the ability to discern between "Lua did not provide the 'source' string" and "The Lua provided 'source' string is not valid utf8"

kyren:
Ehhhh, after reading the Lua source code for the implementation of `lua_getinfo`, I realize that hard-coding the `"nSltu"` parameter is probably a bad idea, because it will be very very costly to do high frequency hooks, more so than I realized at first.  Probably what should happen is that specifying the set of required info should be specified along with the hook in `Lua::set_hook` OR the hook itself should have to call an auxiliary function to specify what information it needs.  The first one is vastly simpler but the second one is more flexible, and also how the normal Lua C API works (probably there is a good reason for this, as well).

Okay, I know that I keep finding things and this is turning into a lot of work for you, and I don't really like that I'm putting so much work onto you.  This is the biggest PR so far to `rlua` and it's for a very core feature and it's becoming a bit unwieldy.

Is there a way that I can do PRs for your PR?  I want to help with all of these features rather than just tell you what to do, as this was something I had fully planned on doing anyway.  Since it's such a core feature, I'm also discovering that the more I think about it the more and more picky I'm becoming.  It would be easier if I could collaborate with code as well as just textual feedback.

BTW I'm not like exactly an expert at open source processes, I'm willing to take suggestions here about how to make things easier.

lemarcuspoilus:
Hello, it's true that you care more and more about the underlying code, but I don't mind. The only downside is that I have other things to do so my progress could be a bit slower than my usual 1-day response.

That being said, feel totally free to make PRs in my PR! Since I'm doing commits directly on my `hooks` branch, you might accidentaly pull invalid code. I don't know how Github fully works but I'll be happy to add write access to your account!

I need to do `&Lua`, `new_ephermal` thing and your suggestion for the `lua_getinfo`'s flags. I suggest you work on the `&str` replacement!

I'll do the stuff later.

lemarcuspoilus:
Actually this PR deserves its own checklist because I'm not even sure what needs to be done :)
I have no idea if you can edit my comment but here's a quick checklist:

- [X] Mutable hook functions
- [X] "lazy" Debug structure to get information
- [ ] better way to handle Lua strings
- [ ] Validate doc and final stuff
- [ ] <insert something here>

kyren:
Couple of quick things:

I feel much better about the `Debug` struct now and getting rid of the hard coded parameter to `lua_getinfo` after your recent changes, this is a lot more reasonable to run with high frequency hooks now and the API seems very reasonable 👍 

`Names`, `Source`, and `Stack` are really generic names, can those be something like `DebugNames`, `DebugSource`, and `DebugStack`?  I'm not picky, and they can be in a submodule if you think that's better but they're a bit too generic to be exported as `rlua::Names`.

There's sort of a proliferation of `pub(crate)`, and I feel a bit bad about having something like `make_ephemeral` be `pub(crate)`, can `hook_proc` just be moved into lua.rs?  I know lua.rs is huge and is a dumping ground, it's a symptom of the crate not having great organization atm, but I don't really want to tackle that right now.  I think doing that removes the need for quite a few `pub(crate)` additions, but I might be missing something.

I think if you push again you'll probably fail travis, because I just recently added rustfmt checks on stable on travis.  I know it's annoying, I'm sorry :(

lemarcuspoilus:
Hello!

> Names, Source, and Stack are really generic names, can those be something like DebugNames, DebugSource, and DebugStack? I'm not picky, and they can be in a submodule if you think that's better but they're a bit too generic to be exported as rlua::Names.

I've heard it was better to avoid prefixing those structures so I left them as-is. And I also preferred to not have to type the submodule before the struct, I find it not as clean. What we can do is re-export them in a submodule in `lib`, if that's feasible.

> There's sort of a proliferation of pub(crate), and I feel a bit bad about having something like make_ephemeral be pub(crate), can hook_proc just be moved into lua.rs? I know lua.rs is huge and is a dumping ground, it's a symptom of the crate not having great organization atm, but I don't really want to tackle that right now. I think doing that removes the need for quite a few pub(crate) additions, but I might be missing something.

I put it there just to make the code more "organized". I don't mind the change but I also like the use of `pub(crate)` in some places (thought you're right I put a lot of them while making the PR). Do what you feel is best of the developers.

> I think if you push again you'll probably fail travis, because I just recently added rustfmt checks on stable on travis. I know it's annoying, I'm sorry :(

It's actually more for you. You know, I like to show a PR with a nice green checkbox right next to it. But to be honest it actually caught one failed build that saved me from letting a bad build stay for too long. I'll keep that in mind.

Btw I'm also sorry for my reduced activity, I have to program on something else before bringing back my attention. Fortunately those are minor details!

kyren:
> I've heard it was better to avoid prefixing those structures so I left them as-is. And I also preferred to not have to type the submodule before the struct, I find it not as clean. What we can do is re-export them in a submodule in lib, if that's feasible.

Well, you could have `pub mod hook` and then only export `HookTriggers` at the top level, that would be okay I suppose, it would just be the first exported submodule.

> I put it there just to make the code more "organized". I don't mind the change but I also like the use of pub(crate) in some places (thought you're right I put a lot of them while making the PR). Do what you feel is best of the developers.

I understand, I'm just trying to put off solving the problem and then solve it "properly", but I don't know what that looks like yet.

> It's actually more for you. You know, I like to show a PR with a nice green checkbox right next to it. But to be honest it actually caught one failed build that saved me from letting a bad build stay for too long. I'll keep that in mind.

What I was trying to say was that before merging you're going to have to format everything with stable rustfmt, hence it would be mildly annoying for you.

> Btw I'm also sorry for my reduced activity, I have to program on something else before bringing back my attention. Fortunately those are minor details!

It's fine, you can work on this at your own pace, there's not a deadline, I'm not going to even release a new version of this crate for a while.

kyren:
> Hello! The suggestions are noted! So actually Travis will check for how I wrote the code? Hmmm I'll satisfy it, hopefully he's not too strict!

It's extremely strict, that's kind of the point.  Don't think about formatting, just run the tool.

> I don't see a format configuration file or something like that either. Technically that means the whole crate would need to be formatted somehow... I might be exaggerating things.

That's exactly how it works, afaik you can just install the rustfmt-preview component and type `cargo fmt`.  It's always been formatted that way, I just never had a travis check until now.

lemarcuspoilus:
Never knew `cargo fmt` existed, I might try that too. I get way too many changes on the preview version. I was using a nightly toolchain, however.

kyren:
> EDIT: Hmmm I'm getting This version of rustfmt is deprecated. Use rustfmt-nightly. Use --force to run deprecated rustfmt even thought the Travis log says nothing... I'm running same Rust version (stable 1.29.1), I guess I'll just force it.

I don't know what version of rustfmt you're using but I believe it is an extremely old version, here is what mine prints:

```
 % rustfmt --version                                                                                                                                                                         
rustfmt 0.99.1-stable (da17b689 2018-08-04)
```

lemarcuspoilus:
Wow, I'm using `0.10.0 ( ) DEPRECATED: use rustfmt-nightly`...
I discovered that, even with the tool uninstalled, it's still available on the command line. I have no idea how it got installed, I guess I'd need to remove it first so I can get a good version.

lemarcuspoilus:
Hello, I don't know if you got stuck too, but running `cargo fmt --all` yields this error (I have the same rustfmt version as you):
```
internal error: left behind trailing whitespace
   --> \\?\E:\rlua\src\conversion.rs:238
```
From some quick searches, this looks like a known bug. I'm not sure how I should tackle this (other than trying the nightly version) without making too many changes to the PR.

By the way, how is the string converter thing doing? I believe we could consider all characters to be ASCII, but that would be limited. As I understood it, it depends on the encoding you're using on the running system. I assume Rust's UTF-8 strings will work just fine but strings returned from a file won't be parsed properly... Would you like me to work on a basic foundation?

kyren:
> Hello, I don't know if you got stuck too, but running cargo fmt --all yields this error (I have the same rustfmt version as you):

I'm not stuck, I was just waiting for you to push a rustfmt update before pushing to your PR.  It's dumb, but lean so heavily on rustfmt that it's mildly painful to edit source where the auto formatting causes edits that I didn't intend.  Normally I would just turn the auto formatting off or manually invoke it for only my edited sections, but hey, this my crate right so I added rustfmt checks to travis instead :D

> internal error: left behind trailing whitespace
   --> \\?\E:\rlua\src\conversion.rs:238

What version of rustfmt are you using, I just tried it on your branch and it works fine?  You didn't even touch `conversion.rs`, so I'm confused.  Can you tell me what `rustfmt --version` at the command line prints?

> By the way, how is the string converter thing doing? I believe we could consider all characters to be ASCII, but that would be limited. As I understood it, it depends on the encoding you're using on the running system. I assume Rust's UTF-8 strings will work just fine but strings returned from a file won't be parsed properly... Would you like me to work on a basic foundation?

`[u8]` is not ASCII, it doesn't specify any encoding, it's just plain bytes.  Lua rules around strings (and identifiers, etc, just code in general) are simplistic, you can shove arbitrary bytes in them and it's not based on the system locale or anything at all.  You can write valid Lua that contains non-utf8 strings and then it's a problem when rust code assumes all strings are utf8.  If all Lua strings were all ASCII that would be fine, because all ASCII is valid utf8 by design.  There is no parsing involved.  You can work on it if you want, but you need to understand what the problem is exactly first.

lemarcuspoilus:
Hello!

> What version of rustfmt are you using, I just tried it on your branch and it works fine? You didn't even touch conversion.rs, so I'm confused. Can you tell me what rustfmt --version at the command line prints?

I'm sorry I was actually not pushing anything. I actually have two separate installations with different behaviors. I'm going to try with `rustfmt 0.99.1-stable (da17b689 2018-08-04)` on the latest stable Rust. Hmmm got the same error. I think I was using nightly on the other station, which was a success but had massive changes everywhere. I have no idea how to fix it unless going full nightly.

For the stable version, I'm using these commands:
```
rustup default stable
rustup component add rustfmt-preview
cargo fmt -- --version
```

That's how I'm getting a few `internal error: left behind trailing whitespace`.

> [u8] is not ASCII, it doesn't specify any encoding, it's just plain bytes. Lua rules around strings (and identifiers, etc, just code in general) are simplistic, you can shove arbitrary bytes in them and it's not based on the system locale or anything at all. You can write valid Lua that contains non-utf8 strings and then it's a problem when rust code assumes all strings are utf8. If all Lua strings were all ASCII that would be fine, because all ASCII is valid utf8 by design. There is no parsing involved. You can work on it if you want, but you need to understand what the problem is exactly first.

Yes I've read about that strings are pretty much just byte arrays. Just got an idea, since _normally_ it should just be some kind of text, could be rely on the existing `String` type of rlua? If so then the changes should be minimal.

Please let me know if you'd like a commit that uses the nightly version! I'll check soon for a reply, I've not been reliable these times.

lemarcuspoilus:
> Hmm, not sure why, but that apparently is only happening on your unpushed changes because it works on the pushed ones. Sorry for making you go through this, the code in your PR before this had a lot of mismatched style, but the whole point of something like rustfmt is not to get into style discussions, so I thought this way would be easier. Clearly it's still not easy though. Just push your changes and don't worry about the travis failure and then I can help you figure out why rustfmt is failing. At the very worst, I can fix up the PR myself either before or after merging.

I do have some small minor style preferences, but I actually don't mind using rustfmt! It's the errors that prevent me from pushing a half-baked commit. As I said nightly works fine but means the PR is going to be a huge PR compatibility destroyer... thought you're the ownerso you do what you want :)

> No, you don't want to do that, because the strings that Lua gives you via the lua_Debug struct are not (or at least they're not presented as) normal internal lua strings, so that would involve copying them. I can fix this to not assume utf8 encoding before or after merging the PR, so just don't worry about it.

I'll leave the code I've originally written as-is for now. I'm still curious to what your solution would actually be, if you can't know for sure how to convert it into a Rust string without failure on many platforms... I'll see when you push the changes.

> No, don't commit with nightly rustfmt, just commit and push so that I can see the error you're getting.

I'm using the latest code on the `hooks` branch, same as you may clone on GitHub right now. I'm doing `cargo fmt --all` right on unmodified sources.

------------------

I read your confession and I totally agree with you! This is the very first time I actually made a PR in a pretty large project without actually knowing what I'm doing. At the same time I feel what you mean by wanting to control it; I'm the same. I like having a nice, neat API that I made and that I fully understand rather than working with someone not thinking the same way and causing inconsistencies.

This, indeed, is close to RFC-class commit (expect I don't have to work in a gigantic code base like it scares me big time) and as time went on I felt your need to tweak the code to your feelings. Those changes are probably more thought out than I would and, subsequently, would be much easier in individual commits.

I'm happy with the code I've written. I think it fits pretty well with the existing API and contains a decent feature set for hooks that should be enough for the average crate user. I never really know if I'll ever be 100% confident with my code but I don't see how to improve it either.

Now, I'd like to go through the paragraphs so I can better structure my thought process.

> so I don't really have the energy to have everything go through a sort of RFC-lite like process, like some more stable projects do.

That discourages me from doing a second, minuscule, PR for another small feature, but I might try anyway. I don't know if there is a PM feature but if so, then I might discuss about it. Nonetheless more details about my motivations below.

> Please don't take this as me saying that your PR is in some way "lacking" or "broken", and I'm "fixing" it, I simply want to iterate on it, which is exactly what I do with features that I write as well. This is turning into an epic RFC level (or, at least it feels like it to me) patch, and I'm not sure I'm ready for this level of formality and process.

I did not take it badly and I don't either :) I expect reviewers to have criticism with the goal to make the source code as best as possible (for the project) since I'll be used for quite some time and probably not changed (in a stable situation thought). My previous PR was accepted right away, no questions asked, which I actually didn't appreciate that much.

I do expect about 50% of the discussion of this PR in subsequent PRs because that's how I feel when a dude comes around with code to give free, but I do not require it either.

> In light of that, what I propose is that I just simply merge your PR in whatever state you're comfortable, and then I make the changes to it I'd like to before releasing the next version of rlua. As long as the PR doesn't introduce unsoundness, I can iterate on the API until I'm happy with it, and if you see me do something that you don't like or would like to improve it in other ways, you can make additional PRs and we can discuss it.

If that wasn't obvious enough I accept you merge it right away (but not without a little comment :D). At the same time I don't see why I should "force" project owners to think my way when I do not own the project or I'm not part of the management group. I tried to mimic the public API that was already existent so the new code feels right at home.

> If you don't like this plan and would like to retain more formal control over the feature that you write, and would like any changes to it to go through PR like discussion, I both completely understand that desire, but also am personally just not ready to give up that level of control yet. This has been a bit awkward for me so far because you seem very invested in this change, which is wonderful and that's a wonderful attitude to have, but I don't want you to give you the wrong impression about the process or lack thereof around this project at this time.

Two parts:

- I feel I gave my fair two cents with the project by defending the `FnMut` choice for the hook function trait requirement and naming of the `HookTriggers` fields, and I can't see myself going much further than that, since I might shown lack of openness.
- I actually have personal interests in using your crate for another of my crazy personal projects. I wanted to modify it to embed it in a `nostd` context but I remembered recently that it would be a bit harder than I thought. The hooks are there to control execution speed and the brief memory suggestion is also related to this. At the same time I didn't want those changes to not be public since they are not specific to such a project.

Thanks too :) I want to give a good impression to the people that I interact with since it makes it easier for me to keep working with them.

> Even if accept your PR then iterate on it, that doesn't mean you haven't still saved me a lot of time and energy exploring the API design space, and that doesn't mean that I'm not extremely grateful.

I'm happy it'll help you go faster with tweaking the API rather than adding extra features! Still surprised, again, since I made it without even knowing how your API really works. I'm also glad that you were open to discuss with me for now an almost 2-week PR! Makes me want to stay here. :)

By the way, I looked up about GCs in general and I now understand why it's a bit harder than `Arc`s all over the place! Haven't looked up your code, I'm still experimenting with some code. Also very tempting to use C++ instead.

kyren:
Hey, so I haven't forgotten about this PR, but I'm possibly about to break it horribly with invasive API changes to fix #97.

Since I was planning on doing some extra work on this when I merged this *anyway*, you don't have to worry about fixing it (unless you just really want to), I'll fix it up when I merge it.  Just letting you know so you don't think I forgot about it.

Also just so you know, I can run `cargo fmt` just fine on your PR, and even did so and pushed it to a local `hooks` branch.  I don't know at all why it's not working for you.  Maybe there is some platform specific rustfmt bug and you and I are on different platforms?

lemarcuspoilus:
Hello! Yes, go ahead and merge it before you have to do massive Git merging nonsense! :) About the formatting, I'm running on Windows 8.1 x64 and I'm editing the files with IntelliJ IDEA and its Rust plugin. I have not tested in a Fedora VM or similar.

I'm going to read #97. By the way, good luck making your changes! We might see each other quite soon with a smaller PR. ;)

PR #96, opened 2018-10-11
Update num-traits dependency to 0.2.6

sakridge:
This crate requires i128 support which is not
added to num-traits until the 0.2.5 release and i128
detection is added in 0.2.6.


kyren:
Oh shoot, good catch!  Thanks very much!

kyren:
I'll make another 0.15 release that includes this fix as that's a pretty bad issue.

sakridge:
Thanks!

PR #98, opened 2018-10-29
Contextify

### LINKED ISSUE #97: Unbounded callback lifetimes cause rlua to be unsound
I'm going to write a very long explanation of this bug, because the root issue here is something that has come up again and again, and I need to document exactly what's going on.

tl;dr I might have an easy fix, but the internal types for callbacks inside rlua are problematic and they probably should change before 1.0. (Edit: I don't have an easy fix D:)

So, the internal type of callbacks in rlua is (basically) this:

```rust
type Callback<'cb> = Box<Fn(&'cb Lua, MultiValue<'cb>) -> Result<MultiValue<'cb> + Send>
```

There are a few different ways to create rust callbacks in rlua, but all of the APIs are basically variations on `Lua::create_function`:

```rust
pub fn create_function<'lua, 'cb, A, R, F>(&'lua self, f: F) -> Result<Function<'lua>>
where
    A: FromLuaMulti<'cb>,
    R: ToLuaMulti<'cb>,
    F: Fn(&'cb Lua, A) -> Result<R> + 'static + Send,
```

These functions take an F and produce the internal `Callback` type, which is then later used to create a Lua callback with `lua_pushcclosure`, but that part isn't terribly important.  What is important here is the `'cb` lifetime: all of the APIs that look like this have the `'cb` lifetime as an unbounded user chosen lifetime, and this is deeply, deeply wrong.  This has always been wrong, and this has been wrong since the very first versions of this crate.

Here's what the type of the internal callback *should* be:

```rust
type Callback = Box<for<'cb> Fn(&'cb Lua, MultiValue<'cb>) -> Result<MultiValue<'cb>> + Send>
```

which is to say that the internal callback should be a function that given *any* `'lua` lifetime, it can take a `Lua` ref and arguments with the same `'lua` lifetime and produce results with that same `'lua` lifetime.  This is the actual, logical requirement that we want to express on the callback, and though we can make this type just fine, problems start when we want to make our callback API.  Let's see how we might change our `create_function` type to allow for the correct callback type:

```rust
pub fn create_function<'lua, A, R, F>(&'lua self, f: F) -> Result<Function<'lua>>
where
    A: for<'cb> FromLuaMulti<'cb>,
    R: for<'cb> ToLuaMulti<'cb>,
    F: for<'cb> Fn(&'cb Lua, A) -> Result<R> + 'static + Send,
```

Except, this will never work, because we have three separate `for<'cb>` HRTBs and there is no way to tell the rust compiler that we need to universally quantify all three trait bounds over a single lifetime.

What's especially frustrating is that it's actually totally possible to *write the code* that produces the correct callback type, but it's not currently possible to put that code into a function and name its type signature; [observe](https://play.rust-lang.org/?gist=ab636f0031f0933a63ac7ac69b330d21&version=stable&backtrace=0).

You can see me struggling with this problem in a very old reddit post [here](https://www.reddit.com/r/rust/comments/5yujt6/a_very_complex_lifetime_problem_possibly_a/).  To proceed with the API and get around this otherwise very tricky or impossible to solve limitation, I picked an API that conveniently "lied" about the callback lifetimes, and I believed this was safe because all callbacks at the time were `'static` anyway.  Well, honestly I had a pretty poor understanding of the borrow checker at the time and I was very unsure of everything, but later on I became slightly more confident that my lie based interface was safe due to the `'static` callback requirement.  I now know this is wrong, and should have realized this sooner.

I lied a bit above, the real callback type inside rlua is actually not what I said above, there is actually a second lifetime parameter to make the callback itself non-'static for use in `Lua::scope`, so the real callback type is actually this:

```rust
type Callback<'cb, 'a> = Box<Fn(&'cb Lua, MultiValue<'cb>) -> Result<MultiValue<'cb>> + Send + 'a>
```

This is to facilitate the `Scope` callbacks having a non-'static lifetime, which if you understand how the argument lifetimes are lying there, you know this is obviously a problem.  At the time I initially made the `Scope` interface I didn't *really* fully understand the implications of the misleading callback types that I was using.  I later found out via #82 that the initial interface that I had written was of course unsound, and it was only after fixing that issue that I more fully understood the problem.

From a certain perspective, the misleading callback types are the root cause of #82, because the callback type requires this user chosen, unbounded `'cb` lifetime and that makes writing a sound interface extremely difficult and easy to get wrong.  I believe the solution for #82 is correct, but it is extremely delicate and intricate and confusing, and this is the LAST property that you want when writing rust code  with `unsafe`.

So the reason that I'm writing about this problem now is not actually due to #82.  Instead it is because I had a minor panic attack today over the following thought: "what if you convinced the borrow checker to pick `'static` as the `'cb` lifetime in rlua callbacks?".

Well, it turns out you can totally do this, and it's obvious that you can do this and I should have realized this sooner.  Observe this monstrosity:

```rust
extern crate rlua;

use std::cell::RefCell;

use rlua::{Lua, Table};

fn main() {
    thread_local! {
        static BAD_TIME: RefCell<Option<Table<'static>>> = RefCell::new(None);
    }

    let lua = Lua::new();

    lua.create_function(|_, table: Table| {
        BAD_TIME.with(|bt| {
            *bt.borrow_mut() = Some(table);
        });
        Ok(())
    }).unwrap()
    .call::<_, ()>(lua.create_table().unwrap())
    .unwrap();

    // In debug, this will panic with a reference leak before getting to the next part but
    // it segfaults anyway.
    drop(lua);

    BAD_TIME.with(|bt| {
        println!(
            "you're gonna have a bad time: {}",
            bt.borrow().as_ref().unwrap().len().unwrap()
        );
    });
}
```

So the *proper* fix is honestly the same as the *proper* fix for #82, which is to fix the callback types, and thus also fix the callback API.  However, without ATCs, it appears that the only way that I can do so is either macros (to write the voldemort function who's type I can't name), or *possibly* some other trait trick I'm not aware of, or possibly redesign `ToLua` / `FromLua` to not require ATCs, I'm not sure.

In the meantime, there's a hopefully simple fix for this issue, but I need to test it out first, and I'll be back in a moment.

kyren:
Fixes #97 and the only current API panic condition

HUGE API change, the entire previous API has been moved to a context callback to take advantage of generative lifetimes.  This both fixed a soundness but and removes an API wart, but requires callbacks to use.


kyren:
There's one thing that was possible before this change that will no longer be possible, and that's mixing outer handles with inner handles from a `Context::scope` callback:

```rust
let lua = Lua::new();
let table = lua.create_table().unwrap();
lua.scope(|scope| {
    scope
        .create_function_mut(|lua, ()| {
            table.set("a", "b").unwrap();
            lua.globals().set("t", table.clone()).unwrap();
            Ok(())
        }).unwrap()
        .call::<_, ()>(())
        .unwrap();
});
assert_eq!(
    lua.globals()
        .get::<_, Table>("t")
        .unwrap()
        .get::<_, String>("a")
        .unwrap(),
    "b"
);
```

Somewhat confusingly, you CAN still do this:

```rust
    Lua::new().context(|lua| {
        let table = lua.create_table().unwrap();
        lua.scope(|scope| {
            scope
                .create_function_mut(|_, ()| {
                    table.set("a", "b").unwrap();
                    Ok(())
                }).unwrap()
                .call::<_, ()>(())
                .unwrap();
        });
        assert_eq!(table.get::<_, String>("a").unwrap(), "b");
    });
```

which is to say, that you can still ofc *access* outer variables inside a scope callback like you would expect, there's just no way to say that the two `'lua` lifetimes are compatible.

You *COULD* imagine a `Context::scope` that actually just re-uses the outer `'lua` lifetime on the inside, but this would not be sound: This would allow `Scope::create_function` to capture `Context<'lua>` from its arguments inside itself, which like everywhere else in rlua is not safe to allow.

The *reason* this is not safe has to do with coroutines.  `Context` is now actually fairly simple, it is just a `ffi::lua_State` pointer and some marker types, nothing else, and this is actually great because it is conceptually much simpler than it was before.  `Context` contains a state pointer, and then marker types with a lifetime that is guaranteed not to outlive the validity of the state pointer it holds.

However, if we used the outer `'lua` lifetime in the inner scoped functions, this would no longer be true!  It would be possible for a script to create a coroutine with its own state pointer, use that coroutine to call the scoped function (which will be passed the state pointer from the coroutine), then garbage collect the coroutine, then call the scoped function again.  If the scoped function could capture `Context`s from its arguments, it would be capturing the invalidated coroutine `lua_State` pointer and that would lead to UB.

This situation would actually be basically the same if I had fixed the unsoundness in #97 in any of the other ways that I mentioned, this is not unique to this particular PR.  The root unsoundness problem is how loose the `'lua` lifetime was which is also what enables this pattern, so fixing one almost has to break the other.

The only alternative that I can come up, which is also potentially conceptually simpler, is to have a single `Context` for the entire duration of `Lua` that is always the main state.  If I did that, it would never actually be harmful to capture arguments, because the `lua_State` pointer lives just as long as `Lua` does (this was mentioned as one of the possibilities to fix #97 as well).  It's a bit gross in that it makes Lua stack usage kind of weird, and it is never a good idea to capture arguments *outside* of a call to `Lua::scope` so that would probably still need to be disallowed, but it would allow this *one* pattern.

Right now, I don't have a huge motivation to do this.  I think this is an *extreme* corner case, and it is almost always a better idea to use `Function::bind` to hold onto outer handle values in a callback. In fact the entire reason I added `Function::bind` was to do this sort of thing -- that way binds will play nicely with the Lua garbage collector, whereas any solution that allows you to capture handles will never play nicely with the gc.  It's only in scoped functions where this isn't as big of a deal since the function has a defined finite lifetime anyway, but even then using `Function::bind` instead might allow the values to be garbage collected *sooner*.

If anybody disagrees with this and thinks that the pattern above is super useful to them, please let me know.  If you think that the alternative lifetime definition makes `Scope` easier to understand, and you think making that change would be worth it, let me know.  Right now I'm I guess mildly against the idea, but I'm willing to discuss it.  I might merge this PR in the next week or so if I can't come up with any additional problems, but even after merging I'm still at least somewhat open to the idea of simplifying the way `Context` works, I just need to know if somebody thinks that's important.

kyren:
I'm going to merge this, I haven't been able to come up with any better ideas and haven't thought of any terrible problems with it.

This still needs more work w.r.t. documentation etc, but it can at least sit in `master` while I iterate on it some more.

PR #104, opened 2018-12-08
Implement ToLua for &[u8]

Hummer12007:
This should be useful for returing arbitrary binary data from functions.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -220,6 +220,12 @@ impl<'lua, 'a> ToLua<'lua> for &'a str {
     }
 }
 
+impl<'lua, 'a> ToLua<'lua> for &'a [u8] {
+    fn to_lua(self, lua: Context<'lua>) -> Result<Value<'lua>> {
+        Ok(Value::String(lua.create_string(self)?))
+    }
+}
+
 macro_rules! lua_convert_int {
     ($x:ty) => {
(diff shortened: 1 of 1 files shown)


kyren:
I'm pretty sure this has been brought up before, but for the life of me I can't seem to find where.  I definitely don't blame you for not seeing the issue since I can't even seem to find it 😛 

I think this is a bit *too* scary and surprising, because `Vec<u8>` and `&[u8]` would be auto-converted entirely differently.  I don't think that's really acceptable, and it's not too onerous to ask the user to call `Context::create_string` in this case to exactly specify what they mean.

I know the boundaries between what implements `ToLua` / `FromLua` are kind of arbitrary, and I admit I don't have a great sense of what definitely should and definitely shouldn't implement those traits.  It's been backed off of before, with things like removing the to / from implementations of `BTreeSet` / `HashSet`.  I think *if* `Vec` implements `ToLua` / `FromLua`, then `&[T]` *cannot* mismatch it, and maybe that `&[T]` *should* implement `ToLua` in that case (which it does not currently), but I'm not sure.

ErichDonGubler:
Does this PR still serve a purpose being open since #140 was merged?

kyren:
> Does this PR still serve a purpose being open since #140 was merged?

No, I think this can be closed, thanks for the reminder!

Issue #106, opened 2018-12-14
Creating a rlua::String from a CString


althonos:
Hi @kyren,

First of all thank you for this extremely useful library, I really enjoy working with it, since it manages to be close to the Lua API while providing a higher level of abstraction.

Yet I've been trying to create a function that returns a `rlua::String`, and couldn't find any way to do it: a `Vec<u8>` ends up being converted as a Table, and of course Rust `String` is not a suitable type to be converted since I'd want the string to contains binary data.

Given the `lua_pushstring` function accepts a `char*`, shouldn't it be possible to implement `ToLua` for `CString` types ? Otherwise, I can't see any way of creating a new `rlua::String` from the Rust code. 

Please do tell me if I'm missing something here ! :smile:

kyren:
Yeah, that seems reasonable to me.  I think `ToLua` should probably be implemented for `CStr` and `CString`?  Should be a simple PR if you'd like to take a stab at it?

PR #107, opened 2018-12-14
Implement `ToLua` for `CStr` and `CString` and `FromLua` for `CString`

althonos:
As discussed in #106 this PR:

* implements `ToLua` for `&CStr` and `CString`
* implements `FromLua` for `CString` (without checking for nul-bytes since Lua guarantees the obtained string is a proper nul-terminated string)

This should offer a safer alternative than simply supporting `&[u8]` as proposed in #104 when working with binary data.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -1,4 +1,5 @@
 use std::collections::{BTreeMap, HashMap};
+use std::ffi::{CStr, CString};
 use std::hash::{BuildHasher, Hash};
 use std::string::String as StdString;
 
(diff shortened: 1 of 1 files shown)


kyren:
> implements FromLua for CString (without checking for nul-bytes since Lua guarantees the obtained string is a proper nul-terminated string

That is *not* true, Lua strings can contain arbitrary data, including embedded nuls.  `String::as_bytes_with_nul` is guaranteed to give you the additional trailing `\0` that Lua adds, but there may also be other `\0` bytes embedded.

althonos:
@kyren: my bad ! I added a safe implementation instead.

PR #108, opened 2018-12-30
Allows loading specific lua modules.

### LINKED ISSUE #51: Loading selected Lua libs
As of right now, when creating a new instance of the Lua structure, all the lua libraries are automatically loaded. This is, as written in the README, undesirable.
Is there any reason this is currently like this? What do I need to know before trying to change the API to make libraries loading into a separate functions?

Shiroy:
Modify the public API to allow the user to load specified subset of the
standard library rather than all the lua standard library. This commit adds 2
new creators to the `Lua` struct:
- A safe `new_with` that force-disable the `debug` library.
- An unsafe `unsafe_new_with` that allows to do everything you want, but does
  not provide any safety guarantees.

Close #51

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ system-lua = ["pkg-config"]
 libc = { version = "0.2" }
 failure = { version = "0.1.2" }
 num-traits = { version = "0.2.6" }
+bitflags = { version = "1.0.4" }
 
 [build-dependencies]
 cc = { version = "1.0", optional = true }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,6 +40,9 @@
 // warnings at all.
 #![doc(test(attr(deny(warnings))))]
 
+#[macro_use]
+extern crate bitflags;
+
 mod error;
 mod ffi;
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -36,6 +36,7 @@ system-lua = ["pkg-config"]
 libc = { version = "0.2" }
 failure = { version = "0.1.2" }
 num-traits = { version = "0.2.6" }
+bitflags = { version = "1.0.4" }
```
Shiroy: This is not strictly necessary but it makes the code neater.

kyren:
This is pretty good, I've been meaning to do this for a while!  I left some review comments, but it's really nothing major.

Sorry for sitting on this for a while before getting to it!

kyren:
Oooone more change and I'll merge it, would it be possible to rename `LuaMod` to something else?  Maybe just `Library` would be fine?  Or `StdLib`?  `LuaMod` is pretty generic and repeats the word "lua", when the rest of the library drops it.

kyren:
Nice, thanks for your work on this!

PR #117, opened 2019-02-04
ptr_to_str takes a c_char not an i8, rm unneeded casts

azdle:
Fixes issue with cross-compiling to armv7, where c_char is a u8.

### DIFF
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -32,8 +32,8 @@ impl<'a> Debug<'a> {
                 "lua_getinfo failed with `n`"
             );
             DebugNames {
-                name: ptr_to_str((*self.ar).name as *const i8),
-                name_what: ptr_to_str((*self.ar).namewhat as *const i8),
+                name: ptr_to_str((*self.ar).name),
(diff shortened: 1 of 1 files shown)


Timidger:
This should probably be changed to `libc::c_char`, which is aliased to whatever the target architecture's type for a c character is (it looks like it's used below as input to the ptr_to_str, it is just the return value that is wrong)

azdle:
@Timidger the fields are already `c_char`. https://github.com/kyren/rlua/blob/dbeb3840520ff0c3467553362b0f0da543a4fbd6/src/ffi.rs#L28

`std::raw::os::c_char` is the same thing as `libc::c_char`. It looks like kyren just changed the types to use the stdlib types and not need the external dep: https://github.com/kyren/rlua/commit/6cfc12313f596489aa54cbb44c3ab21e9e496597

Timidger:
Ah, you are correct. My eyes told me the return type was being cast, but I missed the last `)` encompassing the cast.

kyren:
Yep, that's definitely correct, I missed that when merging #95.

Thanks a lot!

PR #129, opened 2019-04-16
[RFC] add ToLua and FromLua implementations for serde_json::Value

fbenkstein:
This allows passing `serde_json::Value` objects to Lua and returning them.  This serves as a simple bridge between `serde` and `rlua` in cases where the trade-off of going through JSON is acceptable, e.g. when using Lua as a text templating language.

There's currently only a hint of a test.  If the feature itself and the implementation strategy is acceptable I'll try to cover everything with tests.

It might be nice to implement conversions from `serde::Serialize` and to `serde::Deserialize` directly.
Unfortunately, at least to me this looks quite daunting.  I might attempt it later.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,11 +31,13 @@ builtin-lua = ["cc"]
 # the final binary manually.  The builtin-lua and system-lua features are
 # mutually exclusive and enabling both will cause an error at build time.
 system-lua = ["pkg-config"]
+json = ["serde_json"]
 
 [dependencies]
 libc = { version = "0.2" }
 num-traits = { version = "0.2.6" }
 bitflags = { version = "1.0.4" }
+serde_json = { version = "1.0", optional = true }
 
 [build-dependencies]
 cc = { version = "1.0", optional = true }
--- a/src/json.rs
+++ b/src/json.rs
@@ -1,0 +1,108 @@
+use crate::context::Context;
+use crate::error::Error;
+use crate::error::Result;
+use crate::string::String;
+use crate::value::{FromLua, Nil, ToLua, Value};
+
+use serde_json::{Map as JsonMap, Number as JsonNumber, Value as JsonValue};
+
+impl<'lua> ToLua<'lua> for &JsonValue {
+    fn to_lua(self, lua: Context<'lua>) -> Result<Value<'lua>> {
+        Ok(match self {
+            JsonValue::Null => Value::Nil,
+            JsonValue::Bool(b) => Value::Boolean(*b),
+            JsonValue::Number(n) => {
+                if let Some(n) = n.as_i64() {
+                    Value::Integer(n)
+                } else if let Some(n) = n.as_f64() {
+                    Value::Number(n)
+                } else {
+                    Err(Error::ToLuaConversionError {
+                        from: "serde_json::Number",
+                        to: "Integer",
+                        message: Some(format!("value {} too large", n)),
+                    })?
+                }
+            }
+            JsonValue::String(s) => lua.create_string(s).map(Value::String)?,
+            JsonValue::Array(values) => Value::Table(lua.create_sequence_from(values.iter())?),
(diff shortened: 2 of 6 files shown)


fbenkstein:
For some reason, maybe because of `serde_json::Value`'s `From` implementations, this feature breaks type inference in two places in the tests.  I have no idea how this can be avoided or whether this might break users of the crate.  Maybe it's fine if it's not enabled by default.

zrkn:
> It might be nice to implement conversions from serde::Serialize and to serde::Deserialize directly.
Unfortunately, at least to me this looks quite daunting. I might attempt it later.

It was already discussed in https://github.com/kyren/rlua/issues/67
And there is an implementation https://docs.rs/rlua_serde

kyren:
Sorry for not replying to this sooner.

I appreciate wanting the ability to serialize serde_json, and serde Deserialize / Serialize values generally, to / from Lua.

However, I don't *really* want to start supporting external crate types in this crate, but I'm not sure what the best answer is here.  I know that for reasons of coherence, trait impls for `ToLua` and `FromLua` for types from other libraries *have* to be implemented in this crate, but I don't know if I like that solution.

If I add serde_json types, there are a whole *laundry list* of types that probably should be added as well.  So far, I've taken an extremely minimal approach to `FromLua` / `ToLua`, but I could do things differently.

Is there a way to have a minimal rlua core crate and a separate crate for the user-facing `ToLua` / `FromLua` impls, with conversions for lots of different crate types?  Should the core rlua crate not even *have* `ToLua` / `FromLua` traits, and nothing in the core crate is automatically converted?

This is why I've delayed on responding to this PR, because I don't know the best answer here.  I'm interested in other people's thoughts here.

fbenkstein:
Thank you for your response.  I think I can understand your concerns.

First of all, please don't get rid off the `ToLua` / `FromLua` traits.  I think not having them would make `rlua` much harder to use, e.g. when dealing with simple inputs and outputs.  

With regards to `serde`, it and its `Serialize`/`Deserialize` traits are very prominent in the Rust ecosystem.  They are easy to implement  and customize through `serde_derive`. I also think that it's very likely that one would want to use them together.  I just ran a simple grep over the crates.io index and of the 16 crates that mention `rlua` in its entry, 7 also mention `serde`.   I think it would be nice to have `ToLua` and `FromLua` automatically implemented for `Serialize` / `Deserialize` as an optional feature.  Because it's not allowed to implement foreign traits for foreign types this is only possible in the `rlua` crate itself.  For these reasons I think having an optional `serde` feature (probably just taking @zrkn's rlua_serde) in `rlua` could be a very good thing, at least for the users.  However, I can understand if you don't want to take on the additional maintenance burden.

The JSON feature implemented by me in this PR is more of a crutch.  I'd much prefer to use @zrkn's `rlua_serde`.  Unfortunately, I still have two issues with it:

1. it's difficult to discover.  I think mentioning it in the README and/or in the documentation comments for ToLua/FromLua could be very helpful.
2. I cannot currently use it myself because I also need a fix (78c2aac5bda746c9046f701a6d8631ad53841baa) that is not yet in any released version and I could not make `rlua` from git and `rlua_serde` work together.  This will hopefully be resolved as soon as a new version of `rlua` is released.

kyren:
> First of all, please don't get rid off the ToLua / FromLua traits. I think not having them would make rlua much harder to use, e.g. when dealing with simple inputs and outputs.

Oh, I didn't mean that per se, I more meant that there could be an rlua_core crate and an rlua_conversions crate, with an API that provided ToLua / FromLua conversions for a lot of types.

> I cannot currently use it myself because I also need a fix (78c2aac) that is not yet in any released version and I could not make rlua from git and rlua_serde work together. This will hopefully be resolved as soon as a new version of rlua is released.

Well *that* is definitely something that I can fix in the near term, there's no reason not to release a new version with that fix.

Issue #116, opened 2019-02-03
`load` / `loadfile` lua functions can load arbitrary bytecode


kyren:
AIUI this can be used to cause the interpreter to exhibit UB.

`load` and `loadfile` need to be wrapped so that the `mode` flag does not do anything, but unlike pcall / xpcall wrapping, this really seems to break the API of those functions.

There has to be the ability to turn these wrappers off, just like there is the ability to force load the `debug` library, and similarly it will have to be behind an unsafe flag.

The `load `and `loadfile` wrappers shouldn't be too bad, and configuring them could be as simple as adding a flag to `StdLib` that disables wrapping `load` / `loadfile`, though that breaks the `StdLib` <=> 'library' equivalence.

Probably this has to be a new 0.x version bump as well 😞

kyren:
Also `package.loadlib` allows a script to load a C library, which is also really unsafe.

Am I missing any more?

kyren:
Okay, here is a list of everything that I think is potentially unsafe, some of it may simply have to be not loaded *at all*.  Currently only `debug` is guarded against:

- [ ] dofile
- [ ] load
- [ ] loadfile
- [ ] os.setlocale
- [ ] package.loadlib
- [ ] os.date (maybe, might be fixed in 5.2, might be considered a bug)
- [x] debug (okay, technically *some* of debug might be safe, but is of questionable use by itself)

kyren:
It might be simpler to set package.path to "" disallow package.* entirely, as package.path should not be allowed to be set at all.  Without setting package.path to empty string, `require` becomes unsafe.

jugglerchris:
I have spun out #237 (`os.date`) and #238 (`os.setlocale`) as I think they're less serious.
The others are now wrapped or removed by default, with `InitFlags` which can be disabled to allow the originals.

PR #136, opened 2019-06-19
add `load_from_std_lib` function to `Lua` state

azdle:
This function allows loading of any portion of the standard libraries
after initial state creation.

Specifically, this is filling a need to load the debug library after an
error is encountered so that a traceback can be generated without
risking exposing the debug library anywhere else in the code.

### DIFF
--- a/src/lua.rs
+++ b/src/lua.rs
@@ -125,6 +125,42 @@ impl Lua {
         create_lua(lua_mod)
     }
 
+    /// Loads the specified set of safe standard libraries into an existing Lua state.
+    ///
+    /// Use the [`StdLib`] flags to specifiy the libraries you want to load.
+    ///
+    /// Note that the `debug` library can't be loaded using this function as it can be used to break
+    /// the safety guarantees of rlua.  If you really want to load it, use the sister function
+    /// [`Lua::unsafe_load_from_std_lib`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if `lua_mod` contains `StdLib::DEBUG`
+    pub fn load_from_std_lib(&self, lua_mod: StdLib) -> Result<()> {
+        assert!(
+            !lua_mod.contains(StdLib::DEBUG),
+            "The lua debug module can't be loaded using `load_from_std_lib`. Use `unsafe_load_from_std_lib` instead."
+        );
+
+        unsafe {
+            protect_lua_closure(self.main_state, 0, 0, |state| {
+                load_from_std_lib(state, lua_mod);
+            })
+        }
+    }
+
+    /// Loads the specified set of standard libraries into an existing Lua state.
+    ///
+    /// Use the [`StdLib`] flags to specifiy the libraries you want to load.
+    ///
+    /// This function is unsafe because it can be used to load the `debug` library which can be used
+    /// to break the safety guarantees provided by rlua.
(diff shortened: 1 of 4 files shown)


kyren:
Sorry for sitting on this for so long...

I've been trying to figure out exactly what I'm going to do re: #116 for a while, and this PR sort of overlaps with it.

I still need to figure out what I'm going to do, and that will affect everything that loads the Lua stdlib, but I don't think that's a valid reason to wait on this PR, since this is still a strict improvement.

Would it be possible to just combine the body here with the code that loads libraries in `create_lua`?  I'd be happy to merge this then.

azdle:
No worries.

Not sure why I didn't do this in the first place. Force-pushed with the big block of `if`s moved out to a common function.

Also, CI seems to be failing on some not totally backwards compatible change to rust in 1.35 -> 1.36 that's causing rustyline to fail to build. `s/3.0.0/5.0.0/` for rustyline fixes it. Do you want me to just push that in this PR so it builds? Or do you want that change done somewhere else to keep things clean?

kyren:
You can go ahead and push that as part of this PR if you don't mind, I don't mind the drive-by change honestly.

azdle:
There was also a test case failing due to some changes in what errors the compiler produces for broken code. Updated a couple of tests in `tests/compile-fail` to what 1.36 uses too. Hopefully that's right, it looks like only the current stable is built in CI so that should get the shiny green checkmark here.

src/lua.rs
```diff
@@ -513,4 +485,48 @@ unsafe fn create_lua(lua_mod_to_load: StdLib) -> Lua {
     }
 }
 
+unsafe fn load_from_std_lib(state: *mut ffi::lua_State, lua_mod: StdLib) {
+    // Do not open the debug library, it can be used to cause unsafety.
```
kyren: I know you didn't add this, but would you mind removing this comment?

src/lua.rs
```diff
@@ -125,6 +125,18 @@ impl Lua {
         create_lua(lua_mod)
     }
 
+    /// Loads standard the specified set of the standard libraries into an existing Lua state.
+    ///
+    /// Use the [`StdLib`] flags to specifiy the libraries you want to load.
+    ///
+    /// This function is unsafe because it can be used to load the `debug` library which can be used
+    /// to break the safety guarantees provided by rlua.
+    pub unsafe fn load_from_std_lib(&self, lua_mod: StdLib) -> Result<()> {
```
kyren: Could this be `load_std_lib` and also can there be an `unsafe_load_std_lib` variant that allows debug?

azdle: `load_std_lib` implies to me that it loads the whole std lib. Maybe `load_from_std_lib` & `unsafe_load_from_std_lib`? Also any preference on if the first should error if debug is in the list or just ignore it?

kyren: Yeah you're right, I see your point on `load_from_std_lib`, that's fine.

The safe version should *panic* I think if it's asked to load the debug library (that's what new does).

azdle: Ah, should have looked at that. :+1: 

Changed to `load_from_std_lib` and `unsafe_load_from_std_lib`. (Also squashed the fixups.)

azdle:
Had to force-push again because I forgot to `cargo fmt` and I had an extra blank line.

kyren:
Awesome work, thank you very much!

azdle:
Thank *you* for this wonderful library!

PR #140, opened 2019-07-15
add `ToLua` and `FromLua` conversion for `Bytes` and `BytesMut`

azdle:
This is an alternate option to #104 that uses the `bytes`' crate `Bytes` and `BytesMut` types to explicitly mark types that should be interpreted as byte strings rather than arrays of bytes like `Vec<u8>` to hopefully avoid confusion between that an `&[u8]`.

This is more a request for thoughts than anything else. I'm working on a more complete macro crate to derive impls for `ToLua` and `FromLua`, but don't have a good way to handle byte strings since I can't explicitly call `lua.create_string()` and `my_string.as_bytes()` since I don't know the intent of a `Vec` type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ system-lua = ["pkg-config"]
 libc = { version = "0.2" }
 num-traits = { version = "0.2.6" }
 bitflags = { version = "1.0.4" }
+bstr = {version = "0.2", features = ["std"], default_features = false }
 
 [build-dependencies]
 cc = { version = "1.0", optional = true }
--- a/src/context.rs
+++ b/src/context.rs
@@ -841,7 +841,6 @@ impl<'lua> Context<'lua> {
         unsafe {
             let _sg = StackGuard::new(self.state);
             assert_stack(self.state, 1);
-            let source = source.as_ref();
 
             match if let Some(name) = name {
                 ffi::luaL_loadbufferx(
@@ -931,7 +930,7 @@ impl<'lua, 'a> Chunk<'lua, 'a> {
         // "return", then as a statement.  This is the same thing the
         // actual lua repl does.
         let mut expression_source = b"return ".to_vec();
-        expression_source.extend(self.source.as_ref());
+        expression_source.extend(self.source);
         if let Ok(function) =
             self.context
                 .load_chunk(&expression_source, self.name.as_ref(), self.env.clone())
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -3,6 +3,7 @@ use std::ffi::{CStr, CString};
 use std::hash::{BuildHasher, Hash};
 use std::string::String as StdString;
 
+use bstr::{BStr, BString};
 use num_traits::cast;
 
 use crate::context::Context;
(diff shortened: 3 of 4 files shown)


kyren:
I thought about doing this with the types in the `bstr` crate, it seems a bit more appropriate than `bytes` since `bytes` seems pretty specialized to network programming in my mind.

`bstr` seems *precisely* designed for the problem of interacting with things like Lua, but it does have a pretty big scare warning in its documentation about using it in the public interfaces before it hits 1.0.

In general, I am not too interested in maintaining a huge number of `ToLua` / `FromLua` conversions for a huge number of external crates, and I realize that's a bit inconvenient for people, but I just don't have a good sense of what to include and what not to include.

HOWEVER, I think a `bstr` like string type is pretty much going to pass whatever criteria I could possibly come up with, it's pretty clear that a type like this is needed.  And I definitely agree with you that it's especially important when there are proc-macro based conversions involved, because otherwise there is not an owned, 'static type that corresponds to Lua strings without the possibility of failure.

Another possibility is to just include an owned minimal `bstr`-like wrapper type in `rlua` and provide conversion impls for that.  That would save a dependency in a very dependency-light crate, but I don't think the `bstr` crate is a very large dependency.

What do you think?

azdle:
I can definitely understand not wanting the `Bytes` crate as a dependency, that's why I put it behind a feature flag.

`BStr`/`BString` does seem to be pretty much exactly right on what's needed. The warning about never using it in a public interface is worrying, but I'm not sure why it would ever be a problem in normal use.

Currently I'm just using a `ByteString` type defined in my project like this: https://gist.github.com/azdle/2428c4367328c810bdbdc2d421a766ab It covers everything I've needed so far.

I don't really have a hard opinion on which way would be best.

azdle:
Oh, right, my other thought was just leaving the current in-crate interface as is and making a flag in the derive interface like:

```
#[derive(FromLua, ToLua)]
struct Foo {
    #[rlua(bytestr)]
    data: Vec<u8>,
}
```

kyren:
So the choices seem to be:

* Use `bstr`.  I don't really think it's of much use to put it behind a flag, we would just depend on it with no default features I think?
* Make a custom wrapper type with a bare minimum interface like you have in your project.  Would Deref / DerefMut impls to Vec be appropriate here to make it more convenient?
* The status quo, and you work around this problem with proc macro attributes.

If `bstr` didn't have the compatibility scare warning, I would have picked `bstr`, but I think the utility of such a type is greatly diminished if there are conflicting versions of `bstr` in your dependency tree.  I guess option 2 is probably the most convenient and least controversial?  We already have wrapper types like this for things like variadics, so it kind of matches.

Would you be okay with that?

azdle:
I keep going back and forth on what I think would be best. Those are definitely the top three options.

`bstr`: This *feels* the most right, but the warning is scary. I did also have some weird ergonomics problems with the types in here that I didn't expect. That suggests to me that things might still need to change.

Custom Type in rlua: This kinda feels wrong. Mostly because I'd end up with a type from rlua in my external interface or have to convert anyway.

Handle it in attr: Haven't really experimented with this much.

After playing around with things, I'm leaning slightly towards #1 right now. I've pushed that option to this branch. I also played around with #2, that one is commited [here](https://github.com/azdle/rlua/tree/byte-string-convert-bytestring). What used to be on this branch is still available [here](https://github.com/azdle/rlua/tree/byte-string-convert-bytestring), but I agreed that using that doesn't make sense.

So far I wouldn't say I really feel *good* about any of the options.

As I said I haven't played around with #3 yet. If I can find the time I want to do that in the next day or so, otherwise I won't be able to get to it for about a week. Hope it's okay if this PR sits for a bit.

kyren:
Yeah I think this is a good and pretty uncontroversial change, if rlua was a 1.0 library and bstr wasn't I don't think it would be acceptable, but since both are pre-1.0 I think this is fine.  I mean, the only way it could be worse than the status quo is possibly by being slightly more confusing if there were version mismatches with the `bstr` crate, but that downside is really minor.

Are you okay with me merging this now, or would you like to try the proc macro attribute solution?  Even with effortless proc-macro conversion, this *still* seems like it would be an improvement if the user just happened to be using `bstr` anyway.

Like I said before I make really no promises about adding `ToLua` / `FromLua` conversions for other external types, but since there isn't any other owned bytestring type, this is a definite "hole" in the API, and making a custom type to fill it is less convenient and not really any better than this.

So should I merge this?

azdle:
With all that in mind, I think this is ready to merge. As you say it's strictly better and it doesn't preclude a proc macro or a custom type by a user of rlua.

:+1:

kyren:
Thank you!

BurntSushi:
Sorry for the necro bump here, but just saw this as part of my infrequent reverse dependency check.

@azdle 

> I did also have some weird ergonomics problems with the types in here that I didn't expect.

I would very much love to here about these problems! Filing a new issue on `bstr` would be great, even if it's just a brain dump. And of course, given that this was a half-year ago, I understand how this might be difficult.

With respect to the compatibility warning... For the most part, I put that there to discourage folks from _thoughtlessly_ accepting a public dependency on anything that isn't `1.0` yet. This has happened to other crates before and has made them difficult to evolve because too many people were too eager to add it as a public dependency. That's pretty much it.

My _hope_ is that I will get `1.0` out soonish (in the next year?). For the most part, I'm trying to strike the delicate balance of "enough people have used it anger to validate its overall design" vs "too many people are using it to make evolution more difficult." So of course, feedback from y'all would be great!

PR #147, opened 2019-09-15
Improve error message from CallbackError

LPGhatguy:
I got a bug report from a user on my project that uses rlua that the error messages they were getting didn't include the actual error:

https://github.com/rojo-rbx/remodel/issues/7

It turns out `CallbackError` doesn't show its inner error in its `Display` implementation. Is this an appropriate way to signal the inner error in rlua?

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -182,8 +182,8 @@ impl fmt::Display for Error {
             Error::MismatchedRegistryKey => {
                 write!(fmt, "RegistryKey used from different Lua state")
             }
-            Error::CallbackError { ref traceback, .. } => {
-                write!(fmt, "callback error: {}", traceback)
+            Error::CallbackError { ref traceback, ref cause } => {
+                write!(fmt, "callback error: {}: {}", cause, traceback)
             }
             Error::ExternalError(ref err) => write!(fmt, "external error: {}", err),
         }


kyren:
IIRC this comes from a previous project that always printed full cause chains on all errors.  Printing the cause in the `Display` impl of the error itself would cause duplicate error prints, so I avoided it.  Since then rlua has transitioned away from `failure` and the inner error is now printed on`ExternalError` also (and would be pretty weird otherwise), so I think it's probably correct for `CallbackError` to match now.  This will cause errors to be repeatedly printed when printing a full error cause chain, but I honestly don't know how to avoid that in general.

With that in mind, are you okay to merge this?

LPGhatguy:
I think this change is okay to merge. It has a lot of colons in the resulting error message and I'm a little nervous about nested stack traces being presented weirdly, but it's definitely an improvement.

LPGhatguy:
When will the next rlua release be cut? I'd love to have this change for my project deployed to crates.io!

kyren:
I can look at making one sooner rather than later, I was hoping to roll all of the recent fixes in with an update to allow rlua to be used in module position, but I suppose there's no reason to wait.

kyren:
I'm super sorry that I let that languish for so long, I didn't do what I said I'd do.

There were several (very minor) API breaking changes, so rlua 0.17 is now out.

I need to start thinking about getting other maintainers on this repository, since I'm not even currently using Lua or rlua and there probably needs to be somebody else with more of a stake in the crate who can devote more time to it.

PR #152, opened 2019-10-29
Add pairs metamethods

steven-aerts:
Can be used to override the behavior of pairs, for example to iterate
over userData.

http://www.lua.org/manual/5.3/manual.html#pdf-pairs

### DIFF
--- a/src/userdata.rs
+++ b/src/userdata.rs
@@ -63,6 +63,8 @@ pub enum MetaMethod {
     ///
     /// This is not an operator, but will be called by methods such as `tostring` and `print`.
     ToString,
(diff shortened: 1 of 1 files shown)


kyren:
Yeah, that looks correct to me, thanks.

*Technically* this is a breaking change since the `rlua::MetaMethod` enum is not marked as non-exhaustive, but the chances of this breaking anything are very very low, I wonder if it's worth it to just merge this in before releasing 0.16.4

kyren:
Merged, I think I'm going to release a 0.17, sorry for letting this sit for so long!

PR #154, opened 2019-11-09
Added getter for raw source of Lua bytecode chunks

calebwin:


kyren:
Unfortunately, an `rlua::Chunk` is not an actual Lua "chunk", it is just a wrapper around a passed in source string.

You're not the first person to be confused by this, and this is maybe a sign that the type should be renamed.

An actual Lua chunk is more or less indistinguishable from any other Lua closure, and Lua doesn't really give you a lot of control over them once they're compiled.  Thus, the `Chunk` defers compilation into a "real" Lua chunk until you call `Chunk::exec`, `Chunk::eval`, `Chunk::call`, or `Chunk::into_function` on them.

The accessor you're adding would just give you back whatever the input to `Context::load` was.  That's not a terrible thing to add, but it also cannot be unsafe and doesn't need to be boxed or cloned (also, the return value of your PR is just returning a boxed reference, *not* a boxed slice).

calebwin:
I see. Thank you for clearing this up. 👍

kyren:
Just for reference though, you can get the bytecode currently by going through Lua, you can call `string.dump` on any function and it will return bytecode.

PR #164, opened 2020-02-29
Make `Value::type_name` public

a-rodin:
Please make `Value::type_name()` method public.

It would be of great help for users who implement [`FromLua`](https://docs.rs/rlua/0.17.0/rlua/trait.FromLua.html) trait for their custom Rust types and need to create an instance of [`FromLuaConversionError`](https://docs.rs/rlua/0.17.0/rlua/enum.Error.html#variant.FromLuaConversionError) for all of unsupported Lua types at once.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -58,7 +58,8 @@ impl<'lua> Value<'lua> {
             Value::Table(_) => "table",
             Value::Function(_) => "function",
             Value::Thread(_) => "thread",
-            Value::UserData(_) | Value::Error(_) => "userdata",
+            Value::UserData(_) => "userdata",
+            Value::Error(_) => "error",
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kyren:
Sure thing, this isn't exactly related to this change but since this is now going to be public, would you mind removing the space from "light userdata" to "lightuserdata" so all of the types are a single word?  I think Lua prints just "userdata" instead of "lightuserdata" from its lua_typename function (not 100% sure on this), but I think "lightuserdata" is more correct here.

a-rodin:
@kyren I've changed `light userdata` to `lightuserdata`. Also, 91fbc39 shows errors as `error` instead of `userdata`, although errors are technically userdata in Lua, I think the same logic can be applied to them as the one which was applied to light userdata.

PR #165, opened 2020-03-01
Use std::alloc instead of libc::malloc/realloc

### LINKED ISSUE #163: Use std::alloc instead of libc::malloc?
Since now `alloc` is stable, we can use methods from there. `handle_alloc_error` can be used to abort on failure

nazar554:
Fixes #163

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ builtin-lua = ["cc"]
 system-lua = ["pkg-config"]
 
 [dependencies]
-libc = { version = "0.2" }
 num-traits = { version = "0.2.6" }
 bitflags = { version = "1.0.4" }
 bstr = {version = "0.2", features = ["std"], default_features = false }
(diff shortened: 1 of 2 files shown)


kyren:
Thank you for the PR!

I'm going to have to think about this some more before I accept it, because I was unaware that something like [this](https://github.com/kyren/rlua/pull/165/files#diff-00add26f69177c566c48d22452a9dd1dR411-R431) would be required, that is too high of a maintenance burden to keep track of all architectures which rust supports to determine what alignment WOULD have been returned by `libc::malloc`.

I don't know what the right answer here is, but it could be that it's more correct to depend on libc since PUC-Lua also depends on libc and is derived from its behavior?  I'll have to think about it some more.

nazar554:
@kyren The alternative would be to keep the `libc` dependency, but instead call `std::mem::align_of` with `T` of [`libc::max_align_t`](https://en.cppreference.com/w/cpp/types/max_align_t) to get the fundamental alignment

jugglerchris:
@nazar554 Thanks again for the PR!  I'm doing a tidy up, and I think that given the architecture-specific alignment in here that this change isn't worth the extra burden as is so I am closing this PR as is.  (More discussion shortly in #163)

PR #167, opened 2020-03-17
Update CallbackError and ExternalError Display impl to match recommendations

LPGhatguy:
I've learned a fair bit more about Rust error handling since I opened #147. I believe I was now wrong in pushing for the error source being printed in the error's `Display` impl. I'm now trying to correct this in my own projects and projects that I have contributed to in the past. :sweat_smile:

As far as I know, doing "source-chasing" is the responsibility of the code that's printing an error. I think the correct way to handle `CallbackError` and `ExternalError` is as follows:

1. `ExternalError` should forward both `source()` and `Display::fmt` through to the underlying error type. This is what `#[error(transparent)]` does in [thiserror](https://docs.rs/thiserror/1.0.11/thiserror/), but it's easy to do by hand as well.
2. `CallbackError` should do what it did before I filed this PR.

With both of these changes, I think rlua will fit in slightly better into Rust's error ecosystem.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -182,10 +182,10 @@ impl fmt::Display for Error {
             Error::MismatchedRegistryKey => {
                 write!(fmt, "RegistryKey used from different Lua state")
             }
-            Error::CallbackError { ref traceback, ref cause } => {
-                write!(fmt, "callback error: {}: {}", cause, traceback)
+            Error::CallbackError { ref traceback, .. } => {
+                write!(fmt, "callback error: {}", traceback)
             }
-            Error::ExternalError(ref err) => write!(fmt, "external error: {}", err),
+            Error::ExternalError(ref err) => write!(fmt, "{}", err),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


kyren:
Were you running into the same issue I was when I had the error like it was prior to #147, that it doesn't look right when printing a source chain?

So the logic here is: because `ExternalError` is *transparent* it is okay for it to print its source error, but because `CallbackError` contains an error *and* a separate source, it should not print its source?

LPGhatguy:
> Were you running into the same issue I was when I had the error like it was prior to #147, that it doesn't look right when printing a source chain?

When I filed #147, none of my projects were printing source chains and I wasn't aware that was a thing I should've been doing. I started doing that recently and noticed what you did back when I filed that issue!

> So the logic here is: because ExternalError is transparent it is okay for it to print its source error, but because CallbackError contains an error and a separate source, it should not print its source?

Yeah, that lines up with how I think things are supposed to work.

> That would also require changing the impl StdError for Error block, below right?

Oops! Yeah, I think you're right, we need to return our source's source there.

LPGhatguy:
Just pushed up a change to include the `StdError` fix you pointed out.

kyren:
Right, so now nothing prints more than one error in the source chain at a time.  Either the error contains a source error which it returns from `Error::source` and it doesn't print it, or it contains a transparent "cause" error which it does print but it returns the cause's source in `Error::source`.  That makes sense now and seems consistent.

Thanks for following up on this, it had bothered me for a while because I wasn't sure what was appropriate or expected, but now I'm glad we're on the same page.

PR #186, opened 2020-09-06
Updated to Lua 5.4

AlexandraAlter:
A first stab at converting all of this to work on Lua 5.4.

All tests that pass on 5.3 currently pass on 5.4, though three of them were already broken, seemingly due to changes in Rust's compilation errors.
All new features that I've spotted have been opened up for use:
- Multiple user values for full userdata.
- Extra parameter for lua_resume.
- New functions to manage the incremental/generational GC configuration.
- Removal of errors in finalizers - they create warnings now.
- Updated the lua_Debug struct.

Points of uncertainty I'd like input on:
- The means of defining how many user values a UserData type would like. I've added what I think is a decent solution, but I'd like community input.
- Which new functions need to be linked into the library, and how they should be used. Currently I've found warning functions (lua_setwarnf and lua_warning), stack functions (lua_setcstacklimit), to close variables (lua_toclose), GC configuration and threads (lua_resetthread).
- Stack size. Something changed in 5.4 that makes a recursive function overflow the Rust stack, instead of the Lua stack. I've added a const that controls this for now, but would like a more elegant solution. It's probably related to the new function lua_setcstacklimit.

Any other suggestions are welcome!


src/userdata.rs
```diff
@@ -335,32 +340,38 @@ impl<'lua> AnyUserData<'lua> {
 
     /// Sets an associated value to this `AnyUserData`.
     ///
-    /// The value may be any Lua value whatsoever, and can be retrieved with [`get_user_value`].
+    /// The value may be any Lua value whatsoever, and can be retrieved with [`get_i_user_value`].
     ///
-    /// [`get_user_value`]: #method.get_user_value
-    pub fn set_user_value<V: ToLua<'lua>>(&self, v: V) -> Result<()> {
+    /// [`get_i_user_value`]: #method.get_i_user_value
+    pub fn set_i_user_value<V: ToLua<'lua>>(&self, v: V, n: c_int) -> Result<()> {
```
jugglerchris, line 346: Looking at [this comment](https://github.com/amethyst/rlua/issues/172#issuecomment-616975763) from @kyren, I think the hope is to support multiple versions of Lua simultaneously - I guess it would need to be via Cargo features.
If that is desirable, then I wonder whether it's worth keeping `get_user_value()` and `set_user_value()`.  On Lua 5.4 they would just call e.g. `set_i_user_value(v, 1)` and on earlier versions it would call directly.  the `_i_user_value()` methods would be `#cfg` off on < 5.4 (so that you find out at compile time rather than runtime if you're using a feature which doesn't exist in this version of Lua).

jugglerchris:
This all looks pretty good to me, thanks!

There's a bit of a maintenance transition going on at the moment - I'd like to clarify how that's going to work a bit before merging, and give a chance for others to comment.

src/userdata.rs
AlexandraAlter: Ooh, sure. I'll look into that and see if I can get it working over the next few days. It might be a bit complicated by the API changes that have gone on, though, such as the changes in certain function return types.

jugglerchris:
Thanks for your great work and patience!

PR #196, opened 2020-11-02
Impl UserData for `Arc<Mutex<T>> where T: UserData`

gamma-delta:
I'm currently writing a crate so you can `#[derive(UserData)]` on structs, and access their fields from Lua. However, I kept running into borrowing issues. The easiest way to fix those is to use an Arc<Mutex<T>>, but I can't impl UserData for Arc<Mutex<T>> due to the orphan rule. So, I made this fork.

I've included two tests, both of which seem sound. I don't doubt there's still some bugs in the implementation though... There's a lot of `unsafe` code.

The implementation registers every single metamethod, and in the callback, fetches the pointed-to value's metatable, gets the method out of it, and calls it on the pointed-to value. This means, to the Lua user, there should be little difference between using a T and an Arc<Mutex<T>> unless they are doing a lot of `getmetatable`s. The `unsafe` code is mostly `ptr::read` and `ptr::write`, used to copy values out of references and put them back afterwards (aside from the normal unsafety present in the backend).

It currently does *not* overwrite the Metatable method specially.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ pkg-config = { version = "0.3.11", optional = true }
 rustyline = "6.0"
 criterion = "0.3.0"
 compiletest_rs = { version = "0.4", features = ["stable"] }
+crossbeam = "0.8.0"
 
 [[bench]]
 name = "benchmark"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,6 +58,7 @@ mod table;
 mod thread;
 mod types;
 mod userdata;
+mod userdata_wrappers;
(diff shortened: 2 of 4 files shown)


jugglerchris:
Hi,
Thanks for pull request and patience!
I have a couple of questions I've added - possibly because I haven't yet understood the problems.  :-)

jugglerchris (commented):
Hmm, looks like the questions I mentioned above don't appear until I submit the review.  :-\  Sorry about that!

khvzak:
already supported in mlua

PR #200, opened 2020-12-17
derive debug for some rlua types

yjhmelody:
Maybe the low level API's Debug need we to impl it by hand.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -3,7 +3,9 @@
 #![allow(unused)]
 
 use std::mem;
-use std::os::raw::{c_char, c_double, c_int, c_longlong, c_uchar, c_uint, c_ulonglong, c_ushort, c_void};
+use std::os::raw::{
+    c_char, c_double, c_int, c_longlong, c_uchar, c_uint, c_ulonglong, c_ushort, c_void,
+};
 use std::ptr;
 
 pub type lua_Integer = c_longlong;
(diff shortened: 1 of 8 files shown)


jugglerchris:
Hi - do you have a need for `Debug` on those types?  I can't quite see the use case.  :-)

yjhmelody:
I think they're useful when debugging.

src/ffi.rs
```diff
@@ -3,27 +3,36 @@
...
+use std::os::raw::{
+    c_char, c_double, c_int, c_longlong, c_uchar, c_uint, c_ulonglong, c_ushort, c_void,
+};
 use std::ptr;
 
 pub type lua_Integer = c_longlong;
 pub type lua_Unsigned = c_ulonglong;
 pub type lua_Number = c_double;
 
+#[derive(Debug)]
```
jugglerchris, line 15: This seems wrong to me - given that `lua_State` is, as far as the Rust type system is concerned, an empty enum so can't have any instances.

jugglerchris:
Ok, I can't see any particular reason not to have the `Debug` derives except on `lua_State`.  I'm happy to merge if that one is removed.
(In future it would help if rustfmt changes are in separate commits, but no big deal here)

src/ffi.rs
yjhmelody: I will remove it.

jugglerchris:
Thank you!

One comment for future - it's best not to force push to branches in review; it makes it hard to check changes made since the last review.

Issue #145, opened 2019-08-16
Loading and running bytecode


elice02:
Hi,

I have been trying to run some bytecode using this:

let mut f = File::open("C:\\Users\\kkp\\Documents\\Rust\\server_host\\src\\out.luac").unwrap();
let mut buffer = Vec::new();
f.read_to_end(&mut buffer);
lua_ctx.load(&*buffer).exec().is_err();

but I can't get it to work. Do you have any idea how to execute pre-compiled bytecode from a file?

kyren:
The reason for that is that it's potential UB and needs a separate unsafe interface, which I've yet to add.

elice02:
I'm not sure why executing bytecode is considered so insecure? It's part of Lua and is no more or less secure than running arbitrary Lua scripts. Running scripts or bytecode from any insecure resource is a security risk, whether compiled or not. Obviously, scripts allow you to read the content, but that's not really relevant for me. Having precompiled content significantly speeds up what I'm trying to do.

kyren:
It's not that it's insecure, it's that it's actually UB if the bytecode is malformed.  As I understand it, there's no bounds checking on the indexes in the bytecode, so if the bytecode is corrupted or something it will just exhibit UB.

I'm not saying you shouldn't have the ability to do this, I'm just saying it needs a separate unsafe function to do it.

If you really need to do this right now you can do it by going through lua's `loadfile` function instead, which I realize is not super ergonomic.

This is sort of related to #116, I just haven't had time lately to actually sit down and solve all of this.

elice02:
ah ok, I get it. Thanks for the response. I think I will load the file from the file system for now using loadfile. :-)

Many thanks.

PR #206, opened 2021-06-07
Allow loading of binary lua files

allight:
This is useful for some testing applications.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -849,7 +849,7 @@ impl<'lua> Context<'lua> {
                     source.as_ptr() as *const c_char,
                     source.len(),
                     name.as_ptr() as *const c_char,
-                    cstr!("t"),
+                    cstr!("tb"),
                 )
             } else {
                 ffi::luaL_loadbufferx(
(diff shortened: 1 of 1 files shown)


allight:
Failures seem to be the same ones that appear running `cargo test` on master.

jugglerchris:
Hi,
Thanks for the contribution!
However it's deliberate that `load_chunk` doesn't load binary/compiled lua files, so this can't be merged as is.
The intention (which hasn't happened yet) is to add a new, `unsafe` API which does allow loading binary Lua files.  See #116 and #145 for some previous discussion.
I think a method on `Chunk` like:
```rust
pub unsafe fn into_function_unsafe(self) -> Result<Function<'lua>>
```
seems reasonable, but I'd be interested in whether you think that works for your use case.

I'll look into why the CI is failing.

jugglerchris:
FWIW I've fixed the CI issues (which I raised separately as #207).

jugglerchris:
This has been addressed as part of #236.

PR #210, opened 2021-07-17
Add FromLua/ToLua implementation for arrays.

a2aaron:
There is already an existing implementation of FromLua and ToLua for
Vec<T>. This commit adds a similar version of that for arrays. Note that
the implementation checks the Lua array's size and will fail if it does
not exactly match.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -1,4 +1,5 @@
 use std::collections::{BTreeMap, HashMap};
+use std::convert::TryInto;
 use std::ffi::{CStr, CString};
 use std::hash::{BuildHasher, Hash};
 use std::string::String as StdString;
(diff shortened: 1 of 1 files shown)


a2aaron:
Let me know if there's anything I should do to change this PR before it can be merged.

jugglerchris:
Hi, and sorry for the slowness in getting back to you!

The code looks good to me. The const generics means a newish rust is needed, but that's no problem given the time it'll take for us to get to 1.0.

I would ideally like a test adding for the new conversions; if you don't get to it in the next 2-3 weeks I'll do that myself and merge anyway. 

Thanks for your contribution!

jugglerchris:
Merged (with an added test) - thanks again!

PR #227, opened 2021-12-31
feature: lua bindgen

pollend:
this moves the sys binding to another package and lets bindgen produce the bindings. There are some technical problems with versioning I haven't quite figured out and the bindings themselves are mostly correct I just have to re-import them and do some casting to correct some of the constant types. At the moment this is configured through an environment variable `LUA_VERSION`.

I do end up losing functionality i.e I got rid of a way to configure the build through dpkg I think that can easily be fixed just haven't figured out how the build would work though? is it possible to pass environment flags to the sys package?

I was looking at this but it seems to only be available in a nightly build of cargo. 
https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#configurable-env

```
LUA_VERSION=v5.4.3 cargo build
```


jugglerchris:
Hi,
This sounds great, thanks!  I've definitely wanted to move the Lua library into -sys crates, and using bindgen sounds like a great improvement.
I'll try to look properly in the next few days.
Cheers,
Chris

jugglerchris:
I'm definitely in favour of this direction (-sys with bindgen).
Some high level comments:
* I don't think trying to pass environment variables to the -sys crate is a good idea.  Cargo features seem like the obvious way.
* The -sys crate should probably include `rlua` in the same - there is already a `lua-sys` crate on crates.io, so namespacing it seems like a good idea.  (I think it's worth making our -sys crates generic so others can use them if they want, but also keeping them in this project as they're a fundamental part of rlua)
* I think it would be better to have different crates for different major Lua versions - so one each for Lua 5.1, 5.3, 5.4.  Choosing between the included (builtin) vs. system versions should be via a cargo feature.  So the current rlua `builtin-lua53` feature would pull in the `rlua-lua53-sys` crate with the `builtin-lua` feature enabled, but the rlua `system-lua53` feature would also pull in `rlua-lua53-sys` but with the `system-lua` feature enabled instead.

pollend:
aah so like. I guess we would pick the stable version for each submodule.
 
- crates
  - lua53-sys
  - lua51-sys
  - lua54-sys

pollend:
how would the system-lua feature work @jugglerchris? maybe just provide a path through an environment variable would be good?

jugglerchris:
Hi,
I think it should use `pkg-config` as it does now.  I'm not too familiar with bindgen and how easily it can be pointed to system headers.  I can look into that if that helps.

pollend:
> Hi, I think it should use `pkg-config` as it does now. I'm not too familiar with bindgen and how easily it can be pointed to system headers. I can look into that if that helps.

sounds good, just need to point it to the correct include path and run the wrapper.h file.

pollend:
> Hi, I think it should use `pkg-config` as it does now. I'm not too familiar with bindgen and how easily it can be pointed to system headers. I can look into that if that helps.

I think this should work for pkg-config

jugglerchris:
I just had a quick look at the CI failure (on my phone so really only quick). It's complaining about features I think - I don't think this should change - we should just support the same features as before, but with a better internal organisation. 

I'll take a closer look later.

Cargo.toml
```diff
@@ -14,57 +14,28 @@ license = "MIT"
...
-# Builds the correct version of Lua 5.3 inside the crate. If you want to link a
-# specialized version of lua into your binary, you can disable this feature to
-# do that, but care must be taken.  `rlua` makes at least the following
-# assumptions about the linked lua library:
-#   * LUA_INTEGER is long long
-#   * LUA_NUMBER as double
-#   * LUA_EXTRASPACE is at least pointer sized and has at least pointer alignment.
-#   * LUAI_MAXSTACK is 1_000_000
-builtin-lua53 = ["cc"]
+default=[]
```
jugglerchris, line 17: There should still be a default, which should be the builtin Lua 5.4 as before.

Cargo.toml
```diff
@@ -14,57 +14,28 @@ license = "MIT"
...
+rlua_lua54=["lua54_sys"]
```
jugglerchris, line 18: I think the feature names should stay as `builtin-lua54` etc.  But the sys crates should be `rlua-lua54-sys` etc.  They'll want to be published to crates.io eventually.

jugglerchris:
I can see why you changed the feature names, but I think it's best to leave them as is for this PR if that's ok.  I've added a couple of comments on `build.rs` - again I'll try to review the rest soon.  Thanks again for your work!

Cargo.toml
pollend: That's the problem the default features are always enabled. So you end up enabling multiple versions of Lua. Not sure if you can make the selection exclusive.

jugglerchris: You can disable default features (e.g. `cargo build --no-default-features --features builtin-lua53` or `default-features = false` in a dependent crate's Cargo.toml).  The exclusive part is enforced by `build.rs` (it counts the enabled features and errors if != 1).

jugglerchris:
I think it would be better to keep the copy of the Lua source we already have, rather than downloading a new copy.  I'm not keen on downloading things during the build (with the exception of `cargo`'s dependency downloads).

pollend:
> I think it would be better to keep the copy of the Lua source we already have, rather than downloading a new copy. I'm not keen on downloading things during the build (with the exception of `cargo`'s dependency downloads).

hmm, I would opt for a submodule but that is also fine. submodules are pinned to the commit they are associated with. i.e its equivalent to copying in the source from the orignal repository.

crates/lua51_sys/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "lua51_sys"
+version = "0.1.0"
+keywords = ["bindings", "lua"]
+categories = ["external-ffi-bindings"]
+authors = ["Michael Pollind <mpollind@gmail.com>"]
+description = "Bindings to lua's C API"
+repository = "https://github.com/pollend/lua-sys"
```
jugglerchris, line 8: If this is living in the `rlua` repo then this repository link isn't right.

pollend: opps, how would you want the cargo to look like? you can commit to this PR if you have the github cli.

jugglerchris:
Hi,
I've had time to look a bit at the PR (sorry I haven't been able to earlier).
The changes to rlua's Cargo.toml/build.rs look good.

I think the Lua 5.3 and 5.4 sources need to be bundled with the sys crates.  We shouldn't be downloading at build time (once cargo has downloaded dependencies), and the upstream Lua repo is not a source of truth anyway - e.g. the Lua 5.1 tags only go to 5.1.1 (the latest version is 5.1.5).  Ideally the sources should just be an exact copy of the upstream release so that it's easy to verify that it's correct (not actually currently the case with 5.4 :-( ).

While I don't have a strong objection to including a builtin-lua51 option, I think the main reason anyone would to use the Lua 5.1 support rather than 5.3/5.4 is to use an alternative implementation which is only 5.1-compatible (such as LuaJIT).  (Note: I haven't yet tried LuaJIT, but using -sys crates would make that one step closer. :smile: )

I think the -sys crates need to be called something like `rlua-lua53-sys` to avoid confusion with any existing lua*-sys crates on crates.io.  (We probably get away with it as "path" dependencies but...)

jugglerchris:
> hmm, I would opt for a submodule but that is also fine. submodules are pinned to the commit they are associated with. i.e its equivalent to copying in the source from the orignal repository.

I've looked again and it github.com/lua/lua looks fine for 5.4 and 5.3, so that seems fine.  I don't think it has the latest 5.1 versions, but that is ok if we continue to not bother with builtin-lua51...

crates/lua51_sys/Cargo.toml
jugglerchris: Well the repository should point here, and the names need updating as I've said in the other comment.
I've just checked and when publishing to crates.io there can't be any `path =` dependencies when publishing, so these do need to have all the right metadata and will be published as their own crates.

I admit to being a little unsure about what's right for the `authors` field.  I think if it were me doing it I'd leave it out (it's not actually used for much) but I don't feel strongly about it.

pollend:
> > hmm, I would opt for a submodule but that is also fine. submodules are pinned to the commit they are associated with. i.e its equivalent to copying in the source from the orignal repository.
> 
> I've looked again and it github.com/lua/lua looks fine for 5.4 and 5.3, so that seems fine. I don't think it has the latest 5.1 versions, but that is ok if we continue to not bother with builtin-lua51...

was looking at it again and it seems like its an unofficial/official? source but the actual source is through the FTP server.

jugglerchris:
@pollend Thanks for sticking with this!  It'll be 2 or 3 days before I'll have time to come back to this.

src/util.rs
```diff
@@ -818,14 +818,17 @@ where
             ptr::write(ud as *mut WrappedError, WrappedError(err));
             get_error_metatable(state);
             ffi::lua_setmetatable(state, -2);
-            ffi::lua_error(state)
+            ffi::lua_error(state);
+            panic!("code is unreachable")
```
jugglerchris: Minor: this could be `unreachable!()`

crates/rlua-lua51-sys/build.rs
```diff
@@ -0,0 +1,89 @@
+extern crate cc;
+
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    let lua_folder = "lua-5.1.1";
```
jugglerchris, line 7: According to https://www.lua.org/versions.html the latest 5.1 release is 5.1.5.

jugglerchris (approved):
This looks great!  Just a couple of minor comments I'd be happy to fix after merging.
I'll look into what needs doing to fix the CI first before merging.

Thanks again!

jugglerchris:
Merged!  I think I've got a fix for the CI issues in #235 which I'll merge as soon as the CI's done (assuming it passes).

PR #229, opened 2021-12-31
fix: build for stable channel rust

pollend:
this allows the build to work on stable versions of rust. `into_iter()` for array is implemented for `https://doc.rust-lang.org/std/primitive.array.html` 2021 version of rust.

### DIFF
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -392,7 +392,7 @@ lua_convert_float!(f64);
 
 impl<'lua, T: ToLua<'lua>, const N: usize> ToLua<'lua> for [T; N] {
     fn to_lua(self, lua: Context<'lua>) -> Result<Value<'lua>> {
-        Ok(Value::Table(lua.create_sequence_from(vec![self])?))
+        Ok(Value::Table(lua.create_sequence_from(Vec::from(self))?))
     }
 }
 
--- a/src/conversion.rs
+++ b/src/conversion.rs
@@ -392,7 +392,7 @@ lua_convert_float!(f64);
 
 impl<'lua, T: ToLua<'lua>, const N: usize> ToLua<'lua> for [T; N] {
     fn to_lua(self, lua: Context<'lua>) -> Result<Value<'lua>> {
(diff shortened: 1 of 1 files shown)


pollend:
same test fails on master when I test it locally?

jugglerchris:
Hi, and thanks for the fix.
What was your reasoning for this change over the edition update?
I think it's not a problem to update MSRV while we're still 0.x, though it's also nice to support older compilers if there isn't a huge advantage to moving forwards too.

pollend:
I guess vec! creates an allocation on the heap. would be nice to be able to disable sections of the code depending on the version.

jugglerchris:
I haven't had a chance to look properly yet, but doesn't the `vec![self]` just create a vector of arrays which would end up recursively calling the same `ToLua` implementation?

pollend:
> I haven't had a chance to look properly yet, but doesn't the `vec![self]` just create a vector of arrays which would end up recursively calling the same `ToLua` implementation?

I think your right, still learning rust.

pollend:
here are some unit tests that should answer your question: https://github.com/amethyst/rlua/pull/230

jugglerchris:
I've opened a test PR (#231) which passes the tests in CI, so I'm a bit confused about why they're failing for you.
There is a new warning (which I think can break the compile-fail tests) - I'll fix that on my branch.  But I don't see an array into_iter error.

jugglerchris:
I've merged that branch with the macro warning fix - do you still see a problem?

pollend:
> I've merged that branch with the macro warning fix - do you still see a problem?

I guess that answers the question XD https://app.circleci.com/pipelines/github/amethyst/rlua/118/workflows/6f4533f6-6819-4757-8a99-3f0fea5320dd/jobs/172 its overflows the stack and dies.

jugglerchris:
Hi,
Which version of Rust are you seeing the problem on?  master is passing in CI and I've tried locally with 1.56 and 1.57 and not seen any errors or warnings here.

pollend:
seems good :D

PR #230, opened 2021-12-31
chore: added unit testing for primitive conversions

pollend:
here are some primitive conversion unit test.

### DIFF
--- a/tests/conversion.rs
+++ b/tests/conversion.rs
@@ -1,67 +1,66 @@
-use rlua::{Lua, ToLua, Value, Result, Table, Error, String};
+use rlua::{Integer, Lua, Result, String, Table, ToLua, Value};
 
-fn validate_float(verify: Result<Value>, expected: f64) {
+fn valid_float(verify: Result<Value>, expected: f64) {
     let verify_unwrap = verify.unwrap();
     assert_eq!(verify_unwrap.type_name(), "number");
     match verify_unwrap {
         Value::Number(value) => assert_eq!(value, expected),
-        _ => assert!(false, "unexpected type"),
+        _ => panic!("unexpected type"),
     };
 }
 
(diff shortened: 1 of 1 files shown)


tests/conversion.rs
```diff
@@ -0,0 +1,162 @@
+use rlua::{Lua, ToLua, Value, Result, Table, Error, String};
+
+fn validate_float(verify: Result<Value>, expected: f64) {
+    let verify_unwrap = verify.unwrap();
+    assert_eq!(verify_unwrap.type_name(), "number");
+    match verify_unwrap {
+        Value::Number(value) => assert_eq!(value, expected),
+        _ => assert!(false, "unexpected type"),
```
jugglerchris, line 8: This could just be `panic!("unexpected type")`.

jugglerchris:
Thanks for the extra tests!
The CI is failing due to formatting - if you could run `cargo fmt` and rerun, that should fix that part.

jugglerchris:
Looks like there may need to be some `#[any(cfg(rlua_lua53), cfg(rlua_lua54))]` to make the new tests pass on Lua 5.1, which doesn't have integers (all values are floats).

jugglerchris:
I'll add the Lua 5.1 support and merge - I've done a few bits like that so shouldn't take too long.

jugglerchris:
I've added my fixes for Lua 5.1 and opened as #232 - I'll give you a chance to check I haven't done anything silly to your changes before merging.  :-)

pollend:
> I've added my fixes for Lua 5.1 and opened as #232 - I'll give you a chance to check I haven't done anything silly to your changes before merging. :-)

sounds good to me.

jugglerchris:
Thanks for doing this!  It's always good to increase test coverage, and this actually found a problem.

PR #259, opened 2022-08-27
perf: faster HashMap with TypeId keys

Purpzie:
TypeIds are already hashes from the compiler, so they don't need to be hashed again. I've added a dummy hasher which just passes the TypeId through without modifying it.

For example, the `tracing-subscriber` crate does the same thing: https://github.com/tokio-rs/tracing/blob/3517552c319a22c108d423ca9a1675e632f616a3/tracing-subscriber/src/registry/extensions.rs#L12-L34

### DIFF
--- a/src/lua.rs
+++ b/src/lua.rs
@@ -1,6 +1,7 @@
 use std::any::TypeId;
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::hash::{BuildHasherDefault, Hasher};
 use std::marker::PhantomData;
 #[cfg(rlua_lua54)]
 use std::os::raw::c_uint;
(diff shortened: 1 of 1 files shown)


SoniEx2:
relevant: https://github.com/rust-lang/rust/pull/95845

Purpzie:
@SoniEx2 There doesn't seem to be a way to avoid that in the first place, since it's a problem in the compiler itself. Hashing the same inputs still results in the outputs being equal.

SoniEx2:
the point is "stop relying on current implementation details".

like sure, use a faster hasher. but don't break if `TypeId` changes.

jugglerchris:
Hi @Purpzie ,
Thanks for the contribution!
I'm certainly open to performance improvements if they're correct and significant enough to offset any extra complexity.
Have you measured the difference in performance (i.e. is hashing typeids a significant overhead)?

PR #276, opened 2023-06-27
FIX: changes were needed to make repl example work

salexan2001:
I tried to run the example `reple.rs` from the examples folder. Unfortunately it did not work, but yield the following error output:

```
error[E0107]: this struct takes 2 generic arguments but 1 generic argument was supplied
   --> src/main.rs:8:26
    |
8   |         let mut editor = Editor::<()>::new();
    |                          ^^^^^^   -- supplied 1 generic argument
    |                          |
    |                          expected 2 generic arguments
    |
note: struct defined here, with 2 generic parameters: `H`, `I`
   --> (...) /rustyline-12.0.0/src/lib.rs:587:12
    |
587 | pub struct Editor<H: Helper, I: History> {
    |            ^^^^^^ -          -
help: add missing generic argument
    |
8   |         let mut editor = Editor::<(), I>::new();
    |                                     +++

error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> src/main.rs:16:24
   |
16 |                     Ok(input) => line.push_str(&input),
   |                        ^^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `Sized` is not implemented for `str`
   = note: all local variables must have a statically known size
   = help: unsized locals are gated as an unstable feature

error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> src/main.rs:16:21
   |
16 |                     Ok(input) => line.push_str(&input),
   |                     ^^^^^^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `Sized` is not implemented for `str`
note: required by a bound in `Ok`
  --> (...) /library/core/src/result.rs:507:5

Some errors have detailed explanations: E0107, E0277.
For more information about an error, try `rustc --explain E0107`.
error: could not compile `testrlua` due to 3 previous errors 
```

I was able to fix the problems and make the example work again.

Most of the changes displayed in git are whitespace changes due to formatting. In fact the patch is really short and boils down to:

```

4c4
< use rustyline::DefaultEditor;
---
> use rustyline::Editor;
8,11c8
<         // let mut editor = Editor::<()>::new();
<         let rl_c = rustyline::DefaultEditor::new();
<         match rl_c {
<             Ok(mut editor) => {
---
>         let mut editor = Editor::<()>::new();
50,53d46
<         }
< 
<                 },
<             Err(e) => eprintln!("error: {}", e)
```


jugglerchris:
Hi - the example builds for me (and in CI) fine without these changes (and these changes cause the build to fail).
Are you trying to build with a different version of rustyline?

salexan2001:
Yes, that's true. I'm using rustyline version 12.0.0. Sorry, didn't check that.

jugglerchris:
Ah, I'm guessing you've tried to build it in a separate project.
Feel free to bump the version in rlua along with the changes to make it work - I'm happy for it to be more up to date, which is bound to be more useful for anyone trying to use the example as a starting point (as I'm guessing is shown by this MR).

salexan2001:
Tests look better with the bumped version, but unfortunately I have no idea what's wrong with the "build-lua..." tests.

Btw.: You are right, I'm trying to integrate it into a rust-based data analysis tool.

jugglerchris:
I think those failures will go away if you update your branch to the current master - I've recently disabled the bindgen checks as the output doesn't seem so stable recently (e.g. things end up in different orders on different machines).

examples/repl.rs
```diff
@@ -1,49 +1,52 @@
...
-                        ..
-                    }) => {
-                        // continue reading input and append it to `line`
-                        line.push_str("\n"); // separate input lines
-                        prompt = ">> ";
-                    }
-                    Err(e) => {
-                        eprintln!("error: {}", e);
-                        break;
+        // let mut editor = Editor::<()>::new();
```
jugglerchris, line 8: This line can be removed.

jugglerchris:
Great, looks good, thanks - I  just had one trivial comment before merging.

examples/repl.rs
salexan2001: yes, that's right. :-)

jugglerchris:
Excellent, thank you!

PR #280, opened 2023-08-19
Add support for iOS to `rlua-lua54`

w-ensink:
Lua 5.4 supports iOS by defining a dummy `system` function, but `rlua` doesn't support building for iOS because of a missing preprocessor directive. 
This PR doesn't modify any original lua source code, it only adds a compiler flag based on the target platform.

Additionally, I fixed a warning for an unused return value :-)

Probably no need to update `Cargo.toml` to use a relative path after this version is released to crates.io...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ libc = { version = "0.2" }
 num-traits = { version = "0.2.14" }
 bitflags = { version = "1.0.4" }
 bstr = {version = "0.2", features = ["std"], default_features = false }
-rlua-lua54-sys = { path = "crates/rlua-lua54-sys", optional = true }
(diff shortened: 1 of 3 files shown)


jugglerchris:
Looks good, thanks!
Do you mind removing the Cargo.toml change before I merge it?

w-ensink:
> Do you mind removing the Cargo.toml change before I merge it?

Done ✅

jugglerchris:
Thank you!
Do you have any requirements on how quickly a release is made?

w-ensink:
> Do you have any requirements on how quickly a release is made?

Not really, I discovered this while looking to put a Lua interpreter on iOS, but I overlooked the `lua-no-oslib` doing that. I can get without this patch for now by using that feature. So no hurry!

PR #283, opened 2023-09-30
add custom package.path example

paulfrische:
I've added a little example showing how to modify your package.path. This isn't specific to `rlua` but I think many possible users of this crate would find this example helpful.

Have a Great Day!


jugglerchris:
Excellent, thank you!  The CI failures are unrelated - I'll fix those separately.

PR #287, opened 2023-11-18
Implement nested composition of FromLuaMulti

Caellian:
This PR adds ability to compose `FromLuaMulti` values which makes it convenient to add functions which consume arguments of different _kinds_.

This places an additional requirement on user implementing FromLuaMulti to mutate `consumed` argument with the number of arguments that were consumed from provided _remaining arguments_.

I'm not too happy with complexity added by this feature, but it significantly simplifies my workflow in a case where I have to consume arguments in form `(&[(f32, f32)], bool)` and the tuple is supposed to be flattened. The only place that _cares_ about this additional argument is `impl_tuples` trait.
- I decided to add it as a `&mut usize` argument instead of return type as I think that's more convenient for implementers, as it allows gradual mutation of the usage counter instead of having to provide their own. It also provides argument location (didn't test that though).

## Usage

```rust
pub struct FromLuaPoint {
  x: f32,
  y: f32,
}

impl<'lua> FromLuaMulti<'lua> for FromLuaPoint {
  fn from_lua_multi(
        values: LuaMultiValue<'lua>,
        lua: LuaContext<'lua>,
        consumed: &mut usize,
    ) -> LuaResult<Self> {
      match values.next() {
        LuaValue::Number(x) => {
          // consume another...
          *consumed += 2;
          return Ok(FromLuaPoint{ x: x as f32, y: y as f32 })
        }
        LuaValue::Table(table) => {
          // process table...
          *consumed += 1;
          return Ok(FromLuaPoint{ x: table_x, y: table_y })
        }
        _ => {
          // return error when missing arguments, bad arguments, or some other issue...
          return Err(LuaError);
        }
      }
    }
}

methods.add_method("addPoint", |_, this, (point, color): (FromLuaPoint, FromLuaColor)| {
  // do something with point and color parsed from LuaMultiValue
  println!("Added point: {}, {}", point.x, point.y);
});
```

```lua
UserData:addPoint(x, y, color_str)
UserData:addPoint(x, y, r, g, b, a)
UserData:addPoint({x, y}, r, g, b, a)
UserData:addPoint({x = x, y = y}, r, g, b, a)
```

## Breaking changes

`FromLuaMulti::from_lua_multi` takes an additional `consumed: &mut usize` argument.

### DIFF
--- a/src/multi.rs
+++ b/src/multi.rs
@@ -33,8 +33,18 @@ impl<'lua, T: ToLua<'lua>> ToLuaMulti<'lua> for T {
 }
 
 impl<'lua, T: FromLua<'lua>> FromLuaMulti<'lua> for T {
-    fn from_lua_multi(mut values: MultiValue<'lua>, lua: Context<'lua>) -> Result<Self> {
-        Ok(T::from_lua(values.pop_front().unwrap_or(Nil), lua)?)
+    fn from_lua_multi(
+        mut values: MultiValue<'lua>,
+        lua: Context<'lua>,
+        consumed: &mut usize,
+    ) -> Result<Self> {
+        match values.pop_front() {
+            Some(it) => {
+                *consumed += 1;
+                Ok(T::from_lua(it, lua)?)
+            }
+            None => Ok(T::from_lua(Nil, lua)?),
+        }
     }
 }
 
@@ -45,7 +55,12 @@ impl<'lua> ToLuaMulti<'lua> for MultiValue<'lua> {
 }
 
 impl<'lua> FromLuaMulti<'lua> for MultiValue<'lua> {
-    fn from_lua_multi(values: MultiValue<'lua>, _: Context<'lua>) -> Result<Self> {
+    fn from_lua_multi(
+        values: MultiValue<'lua>,
+        _: Context<'lua>,
+        consumed: &mut usize,
+    ) -> Result<Self> {
+        *consumed += values.len();
         Ok(values)
     }
 }
@@ -84,6 +99,11 @@ impl<T> Variadic<T> {
     /// Creates an empty `Variadic` wrapper containing no values.
     pub fn new() -> Variadic<T> {
         Variadic(Vec::new())
+    }
+
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.0.len()
     }
 }
 
@@ -129,12 +149,20 @@ impl<'lua, T: ToLua<'lua>> ToLuaMulti<'lua> for Variadic<T> {
 }
 
 impl<'lua, T: FromLua<'lua>> FromLuaMulti<'lua> for Variadic<T> {
-    fn from_lua_multi(values: MultiValue<'lua>, lua: Context<'lua>) -> Result<Self> {
-        values
+    fn from_lua_multi(
+        values: MultiValue<'lua>,
+        lua: Context<'lua>,
+        consumed: &mut usize,
+    ) -> Result<Self> {
+        let result = values
             .into_iter()
             .map(|e| T::from_lua(e, lua))
             .collect::<Result<Vec<T>>>()
-            .map(Variadic)
+            .map(Variadic);
+        if let Ok(result) = &result {
+            *consumed += result.len();
+        }
+        result
     }
 }
 
@@ -147,38 +175,48 @@ macro_rules! impl_tuple {
         }
 
         impl<'lua> FromLuaMulti<'lua> for () {
-            fn from_lua_multi(_: MultiValue, _: Context<'lua>) -> Result<Self> {
+            fn from_lua_multi(_: MultiValue, _: Context<'lua>, _: &mut usize) -> Result<Self> {
                 Ok(())
             }
         }
     );
 
-    ($last:ident $($name:ident)*) => (
-        impl<'lua, $($name,)* $last> ToLuaMulti<'lua> for ($($name,)* $last,)
-            where $($name: ToLua<'lua>,)*
-                  $last: ToLuaMulti<'lua>
+    ($($name:ident)+) => (
+        impl<'lua, $($name),*> ToLuaMulti<'lua> for ($($name,)*)
+            where $($name: ToLuaMulti<'lua>),*
         {
             #[allow(unused_mut)]
             #[allow(non_snake_case)]
             fn to_lua_multi(self, lua: Context<'lua>) -> Result<MultiValue<'lua>> {
-                let ($($name,)* $last,) = self;
+                let ($($name,)*) = self;
 
-                let mut results = $last.to_lua_multi(lua)?;
-                push_reverse!(results, $($name.to_lua(lua)?,)*);
+                let mut results = MultiValue::new();
+                push_reverse!(results, $($name.to_lua_multi(lua)?,)*);
                 Ok(results)
             }
         }
 
-        impl<'lua, $($name,)* $last> FromLuaMulti<'lua> for ($($name,)* $last,)
-            where $($name: FromLua<'lua>,)*
-                  $last: FromLuaMulti<'lua>
+        impl<'lua, $($name),*> FromLuaMulti<'lua> for ($($name,)*)
+            where $($name: FromLuaMulti<'lua>),*
         {
             #[allow(unused_mut)]
             #[allow(non_snake_case)]
-            fn from_lua_multi(mut values: MultiValue<'lua>, lua: Context<'lua>) -> Result<Self> {
-                $(let $name = values.pop_front().unwrap_or(Nil);)*
-                let $last = FromLuaMulti::from_lua_multi(values, lua)?;
-                Ok(($(FromLua::from_lua($name, lua)?,)* $last,))
+            fn from_lua_multi(mut values: MultiValue<'lua>, lua: Context<'lua>, total: &mut usize) -> Result<Self> {
+                $(
+                    let $name = {
+                        let mut consumed = 0;
+                        let it = match FromLuaMulti::from_lua_multi(values.clone(), lua, &mut consumed) {
+                            Ok(it) => it,
+                            Err(err) => {
+                                return Err(err);
+                            }
+                        };
+                        *total += consumed;
+                        values.drop_front(consumed);
+                        it
+                    };
+                )*
+                Ok(($($name,)*))
             }
         }
     );
@@ -187,11 +225,11 @@ macro_rules! impl_tuple {
 macro_rules! push_reverse {
     ($multi_value:expr, $first:expr, $($rest:expr,)*) => (
         push_reverse!($multi_value, $($rest,)*);
-        $multi_value.push_front($first);
+        $multi_value.push_front_many($first);
     );
 
     ($multi_value:expr, $first:expr) => (
-        $multi_value.push_front($first);
+        $multi_value.push_front_many($first);
     );
 
     ($multi_value:expr,) => ();
--- a/src/value.rs
+++ b/src/value.rs
@@ -142,8 +142,16 @@ impl<'lua> MultiValue<'lua> {
         self.0.push(value);
     }
 
+    pub(crate) fn push_front_many(&mut self, values: Self) {
+        self.0.extend(values.into_iter());
+    }
+
     pub(crate) fn pop_front(&mut self) -> Option<Value<'lua>> {
         self.0.pop()
+    }
+
+    pub(crate) fn drop_front(&mut self, count: usize) {
+        self.0.drain(self.0.len() - count..).count();
     }
 
     pub fn len(&self) -> usize {
@@ -179,5 +187,9 @@ pub trait FromLuaMulti<'lua>: Sized {
     /// values should be ignored. This reflects the semantics of Lua when calling a function or
     /// assigning values. Similarly, if not enough values are given, conversions should assume that
     /// any missing values are nil.
-    fn from_lua_multi(values: MultiValue<'lua>, lua: Context<'lua>) -> Result<Self>;
+    fn from_lua_multi(
+        values: MultiValue<'lua>,
+        lua: Context<'lua>,
+        consumed: &mut usize,
+    ) -> Result<Self>;
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -179,7 +179,8 @@ impl<'lua> Context<'lua> {
         F: 'static + Send + Fn(Context<'lua>, A) -> Result<R>,
     {
         self.create_callback(Box::new(move |lua, args| {
-            func(lua, A::from_lua_multi(args, lua)?)?.to_lua_multi(lua)
+            let mut consumed = 0;
+            func(lua, A::from_lua_multi(args, lua, &mut consumed)?)?.to_lua_multi(lua)
         }))
     }
 
@@ -365,8 +366,12 @@ impl<'lua> Context<'lua> {
     }
 
     /// Converts a `MultiValue` instance into a value that implements `FromLuaMulti`.
-    pub fn unpack_multi<T: FromLuaMulti<'lua>>(self, value: MultiValue<'lua>) -> Result<T> {
-        T::from_lua_multi(value, self)
+    pub fn unpack_multi<T: FromLuaMulti<'lua>>(
+        self,
+        value: MultiValue<'lua>,
+        consumed: &mut usize,
+    ) -> Result<T> {
+        T::from_lua_multi(value, self, consumed)
     }
 
     /// Set a value in the Lua registry based on a string name.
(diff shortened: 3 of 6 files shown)


src/multi.rs
```diff
@@ -147,38 +175,48 @@ macro_rules! impl_tuple {
...
 
-                let mut results = $last.to_lua_multi(lua)?;
-                push_reverse!(results, $($name.to_lua(lua)?,)*);
+                let mut results = MultiValue::new();
+                push_reverse!(results, $($name.to_lua_multi(lua)?,)*);
                 Ok(results)
             }
         }
 
-        impl<'lua, $($name,)* $last> FromLuaMulti<'lua> for ($($name,)* $last,)
```
Caellian, line 172: Please check this change. I'm guessing `$last` was separate here to allow for `Variadic` last argument.
- `Variadic` can now be in any location but will swallow and error at runtime for inconvertible `Value`s.
  - Should probably be fixed so it works similar to `ab*?c` regex.
- Was the fact that `A` value was placed last in any other way significant?

jugglerchris:
Hi,
Thanks for the merge request!

I wonder if this is maybe a bit too much extra complexity for what seems like is an unusual case, plus I worry that there's a lot of scope for confusion when some type consumes more (or fewer) values than the user expected.  In my experience it can sometimes be tricky debugging issues where the automatic argument conversions fail (your function ends up not being called), and this feels like they'd have more ways to go wrong.

Do you gain much from this modified trait being in `rlua`, compared to having it in your application?  Your function could just take a `MultiValue<'lua>`, and your types can have a `construct_from_multi(&mut MultiValue<'lua>, &mut consumed)` which you call for both (for example) types?

Caellian:
> Do you gain much from this modified trait being in rlua, compared to having it in your application?

I register 50+ methods which all take `Point` types and want to allow passing it as a Table from Lua so that they can be passed around/generated easily. Unpacking `MultiValue` in all those places would make my code a lot more complex because I'd have to move the `consumed` counter into each one of those.

A simple case [here](https://github.com/Caellian/clunky/blob/2b20126137c267acf82807567b01eba3f04777eb/src/skia_bindings.rs#L1042) would add like ~5 more lines because I'd have to match against the tailing `f32` manually after calling `construct_from_multi`.

It also makes it very straightforward to implement functions with `Variadic` length that have tailing arguments ([here](https://github.com/Caellian/clunky/blob/2b20126137c267acf82807567b01eba3f04777eb/src/skia_bindings.rs#L1499C2-L1499C2)).
- I plan on replacing that `Variadic` with a `Table` to closer match the Skia API so not the best use case example.
- This works like `a(bc)*d` regex, so the error will be on `d` if either `b` or `bc` capture fails.

In [some cases](https://github.com/Caellian/clunky/blob/2b20126137c267acf82807567b01eba3f04777eb/src/skia_bindings.rs#L1695C5-L1695C5) I have 5 arguments all of which would have to be manually type checked. And in the linked one, while the argument isn't actually a point, unpacked form makes sense as well as being able to provide them as a table.

I also [sometimes](https://github.com/Caellian/clunky/blob/2b20126137c267acf82807567b01eba3f04777eb/src/skia_bindings.rs#L2214) need to optionally capture either 2 (x & y) arguments or none.
- In case `Option<Multi>` is in the middle it works like `a(bc)?d` regex, error is on `d`.

This change makes it very convenient to [configure the number of consumed arguments](https://github.com/Caellian/clunky/blob/2b20126137c267acf82807567b01eba3f04777eb/src/skia_bindings.rs#L1257) in method signature instead of body which I find nicer, but same functionality could be achieved with proposed `construct_from_multi`.

I considered what you're suggesting at first, as well as forcing separate x & y arguments or a table, but I think this provides best experience from Lua code while at the same time making the binding code much cleaner. Implementation of `FromLuaMulti` stays mostly the same except for having to [update the counter](https://github.com/Caellian/clunky/blob/2b20126137c267acf82807567b01eba3f04777eb/src/skia_bindings.rs#L623).
- [ ] I'll have to add that to the `FromLuaMulti` doc comment.

I could provide index of failed top-level argument Type (and corresponding Lua argument index) in the error message if you find it necessary, but from what I experimented with 1-level of nesting, the error messages produced from `LuaPoint` type seemed enough for me to debug issues while it had bugs.
- It would be nice though if error were to say: "value X doesn't continue either Y nor is valid for Z: _Z inner message_"
  - I'd have to add and check for a new error type that'd be returned by `Option` and `Variadic`, but it's doable.

Caellian:
- [ ] CI failing reminds me to add tests for these features.

jugglerchris:
Sorry for the delay between replies!

I can see that this is a very useful feature for some use cases!  To recap, my current concerns are:

* Extra complexity and chance of errors, for example:
  * The consumed count is another thing to have to get right (and very slightly more annoying to implement for existing cases)
  * There are more ways to have confusing behaviour when the FromMulti is not last, e.g. values not being consumed by the expected parameter,
* It's a breaking change.  That in itself is fine, we can bump the version number, but I'm concerned that having to update all the users' FromLuaMulti impls to add the right counts makes it a fairly annoying update.

So I'd like to make sure we can balance the problems with the benefits.

Some thoughts I've had (but not had the time to investigate properly):

1. Can this be done by adding another wrapper type that implements the original `FromLuaMulti`, but that delegates to your modified version, without changing the original?  Something like:

```rust
pub struct FlexibleFromLuaMulti<T>(T);

... generated from macro for different sized tuples
impl<A:NewFromLuaMulti, B:NewFromLuaMulti> FromLuaMulti for FlexibleFromLuaMulti<(A, B)> {
    fn from_lua_multi(...) {
        let mut consumed = 0usize;
        // delegate to NewFromLuaMulti::from_lua_multi(..., &mut consumed)
    }
}

// using it
methods.add_method("addPoint", |_, this, FlexibleFromLuaMulti((point, color)): FlexibleFromLuaMulti<(FromLuaPoint, FromLuaColor)>| {
  // do something with point and color parsed from LuaMultiValue
  println!("Added point: {}, {}", point.x, point.y);
});
```
I guess a shorter name would probably be needed!

2. As a different option, is there a way to do this by *adding* the method to `FromLuaMulti` which lets simpler implementations (i.e. most of them) keep their current implementation?  I'm thinking of adding the new method with the count parameter, but with a default implementation which calls the old version and increments the count by the number of values (i.e. consuming all of them).

The (probably fatal, but maybe there's a creative solution) issue I can see here is that I don't think there's a way to do this without either forcing some users to implement both methods, or having an impl which implements neither which compiles but fails at runtime.


I'm still thinking about this - I'd appreciate your thoughts.

Caellian:
Right, a solution that's backwards compatible did cross my mind at some point but I didn't have time to implement it and half-way forgot about it, but it was something like:
```rust
pub trait FromArgPack<'lua> {
  // current FromLuaMulti::from_lua_multi
}

impl<'lua, M: FromLuaMulti<'lua>> FromArgPack for M {
  // pass to multi without counter
}
// `impl for FromLua` is done though `impl FromLuaMulti for FromLua`

// use `FromArgPack` instead of `FromLuaMulti` in `impl_tuple`
```

I think that would:
- keep the old behavior and API untouched,
  - besides now allowing `Variadic` in the middle,
- isolate the complexity from `FromLuaMulti`,
- avoid having to wrap arguments in another type, i.e. look cleaner,
- avoid having to clone `MultiValue` for each try,
  - might require specialization for `FromLua` though, so probably _will_ clone
- provide `impl_tuple` more info and make the macro itself smaller (less generated code).

jugglerchris:
Hi,
That idea sounds really good to me.

>  I didn't have time to implement it and half-way forgot about it

Do you think it's still a lot of work to get to that from where you are now?  I may find time to pitch in a bit.

Caellian:
> Do you think it's still a lot of work to get to that from where you are now? I may find time to pitch in a bit.

Shouldn't be a lot of work, just have to replace `FromLuaMulti` with a differently named trait in most cases. I'm finishing up my Skia bindings atm, once I'm finished with that I'll get to it - so this or the following weekend.

src/value.rs
```diff
@@ -170,14 +182,45 @@ pub trait ToLuaMulti<'lua> {
...
-    /// assigning values. Similarly, if not enough values are given, conversions should assume that
-    /// any missing values are nil.
-    fn from_lua_multi(values: MultiValue<'lua>, lua: Context<'lua>) -> Result<Self>;
+    /// In case `values` contains more values than needed to perform the
+    /// conversion, the excess values should be ignored. This reflects the
+    /// semantics of Lua when calling a function or assigning values. Similarly,
+    /// if not enough values are given, conversions should assume that any
+    /// missing values are nil.
+    fn from_lua_multi(values: MultiValue<'lua>, lua: Context<'lua>) -> Result<Self> {
+        Self::from_counted_multi(values, lua, &mut 0)
```
Caellian, line 200: I think adding a separate method like this is probably the cleanest solution that would case least breakage in dependent code. It also makes the diff a lot more concise.

The only issue I can see is someone implementing `FromLuaMulti` without overriding either method which would cause infinite recursion at runtime, but I think it's obvious that _something_ has to be specified (and it's noted in the doc).

Caellian: The benefit of having those methods call each other is that dependents only need to implement one, depending on whether they need to handle arguments non-exhaustively.

Caellian:
@jugglerchris I think this is the final solution, just let me know whether you think there would be any benefit to passing the `MultiValue` by reference. I'm not sure because I noticed `LuaRef` has some custom logic for reference counting and I'm assuming it's not a huge cost to clone them but can't really tell without getting more familiar with inner workings of the library.

src/multi.rs
```diff
@@ -128,57 +148,154 @@ impl<'lua, T: ToLua<'lua>> ToLuaMulti<'lua> for Variadic<T> {
...
+
+impl<'lua, T: FromLuaMulti<'lua>> FromLuaMulti<'lua> for Fallible<T> {
+    fn from_counted_multi(
+        values: MultiValue<'lua>,
+        lua: Context<'lua>,
+        consumed: &mut usize,
+    ) -> Result<Self> {
+        match T::from_counted_multi(values, lua, consumed) {
+            Ok(it) => {
+                *consumed += 1;
```
jugglerchris, line 252: What is this increment for?  I would expect the `T::from_counted_multi()` would already have done this.

src/value.rs
```diff
@@ -142,10 +142,22 @@ impl<'lua> MultiValue<'lua> {
         self.0.push(value);
     }
 
+    pub(crate) fn append(&mut self, values: Self) {
```
jugglerchris, line 145: Minor: as an occasional Python user I assumed append added one value.  Maybe `extend` would be better?

src/value.rs
```diff
@@ -170,14 +182,45 @@ pub trait ToLuaMulti<'lua> {
...
+    /// Failing to do so will make the outer scope try and use the same sequence
+    /// of `values` for multiple consecutive conversions. In case of an error,
+    /// the `consumed` counter value should be kept the same to allow wrappers
+    /// such as [`Variadic`] to conservatively capture values.
+    ///
+    /// [`Variadic`]: struct.Variadic.html
+    fn from_counted_multi(
+        values: MultiValue<'lua>,
+        lua: Context<'lua>,
+        consumed: &mut usize,
```
jugglerchris, line 219: Rather than incrementing consumed, should it just start at 0 and the implementer just writes the number consumed, rather than incrementing a value they're not interested in?

src/conversion.rs
```diff
@@ -440,10 +441,27 @@ impl<'lua, T: ToLua<'lua>> ToLua<'lua> for Vec<T> {
...
+            let mut values: MultiValue = MultiValue::from_vec(
+                table
+                    .sequence_values::<Value>()
+                    .filter_map(Result::ok) // Every value in Table is always a Value
+                    .collect(),
+            );
+            let mut result = Vec::new();
+            while values.len() > 0 {
+                let mut consumed = 0;
+                match T::from_counted_multi(values.clone(), lua, &mut consumed) {
```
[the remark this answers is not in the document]
Caellian: The main drawback with that approach is that implementations have to manually juggle values (revert them when testing fails). `Fallible` could do a clone and revert it on error but it complicates the implementation for most other structures.

I tried adding a wrapper as well at some point but it required adding lifetimes to all the traits/functions that deal with `FromLuaMulti` conversion which felt like it complicates code too much. Though I do think an iterator with "revert last argument" function and "list skipped types" function would be ideal.

src/multi.rs
Caellian: Right, that's an error. Didn't test this latest version as much.

src/value.rs
Caellian: In rust the following naming convention is used in stdlib:
- [push](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push) - single owned value
- [append](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) - push all values from `&mut Self`
- [extend](https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend) - with values returned from `Iterator`

The argument should be a `&mut Self` though.

src/value.rs
Caellian: That could also work but it makes the code more complicated as consumed would have to be constructed from scratch before each call. In case of nested arguments that are composed out of several others, that's a lot more room for error.

src/value.rs
Caellian: I think this is the best solution possible atm. Alternatives are:
- depending on nightly,
- having to manually implement both methods, or
- the first solution which requires possibly a lot of updates to existing code base.

Caellian:
While reading your review, I thought about maybe removing the added method and changing the existing `from_lua_multi` to:
```rust
fn from_lua_multi(values: &mut MultiValue<'lua>, lua: Context<'lua>) -> Result<Self>
```
which is a mix of one of your suggestions and the initial version.

It's good because:
- it's the simplest solution that handles counting automatically,
- avoids the infinite recursion issue when trait is incorrectly implemented,

but the downsides are:
- it's a breaking change,
- requires manually reverting values in some (infrequent) cases.

I think we've covered all the different ways this could be implemented currently. Additionally, I could add a feature that enables specialization and provides the ideal API if the user doesn't mind using the nightly compiler, though it's up to you to pick which stable tradeoffs you find acceptable.

I think this last suggestion to change `from_lua_multi` to take a `&mut MultiValue` is the cleanest and avoids most pitfalls. I would also accompany this with changing `from_lua` to take a reference as well for consistency.

src/conversion.rs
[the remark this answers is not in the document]
jugglerchris: I agree that adding extra lifetimes to traits would be awkward, and make updates from previous versions difficult.
But I think the quadratic runtime for long (even non-multi-value) argument lists is also a problem.
That quadratic issue is only because multiple arguments can have a crack a (nearly) the same list, and wouldn't be a problem with the old "take all" version.  Maybe the new counted method can take a different type which does behave as an iterator.  Perhaps with a "peek next" rather than "revert pop"?

src/value.rs
jugglerchris: Ok, fair enough!

jugglerchris:
Sorry again for slow replies!

> While reading your review, I thought about maybe removing the added method and changing the existing `from_lua_multi` to:
> 
> ```rust
> fn from_lua_multi(values: &mut MultiValue<'lua>, lua: Context<'lua>) -> Result<Self>
> ```
> 
> which is a mix of one of your suggestions and the initial version.
> 
> It's good because:
> 
>     * it's the simplest solution that handles counting automatically,
> 
>     * avoids the infinite recursion issue when trait is incorrectly implemented,
> 
> 
> but the downsides are:
> 
>     * it's a breaking change,
> 
>     * requires manually reverting values in some (infrequent) cases.

I think changing the `&mut` is a reasonable breaking change.  The contents of the implemented methods would mostly not change - occasionally there might need to be a small change in places where the `MultiValue` was moved.

> I think this last suggestion to change `from_lua_multi` to take a `&mut MultiValue` is the cleanest and avoids most pitfalls. I would also accompany this with changing `from_lua` to take a reference as well for consistency.

I don't see any reason to change `from_lua` though - `Value` is already really a kind of reference, and I don't see a benefit to breaking existing impls here.

Caellian:
I added some small utility functions to `MultiValue` as well so it's easier to pop and revert values from implementations.

src/value.rs
```diff
@@ -170,14 +186,19 @@ pub trait ToLuaMulti<'lua> {
 
 /// Trait for types that can be created from an arbitrary number of Lua values.
 ///
-/// This is a generalization of `FromLua`, allowing an arbitrary number of Lua values to participate
-/// in the conversion. Any type that implements `FromLua` will automatically implement this trait.
+/// This is a generalization of `FromLua`, allowing an arbitrary number of Lua
+/// values to participate in the conversion. Any type that implements `FromLua`
+/// will automatically implement this trait.
+///
+/// Implementing either `from_lua_multi` or `from_counted_multi` is required as
```
jugglerchris, line 193: The comment about `from_counted_multi` is now out of date.

jugglerchris (commented):
This is looking great, thanks!  I just noticed one out of date comment, otherwise I think this is good to merge.  Thanks for your patience!

Caellian:
Give me a few days to update my binding code and maybe add some tests. There have been a few minor semantic changes I did since I last properly tested the entire modified surface area so I'd just like to make sure stuff like `Variadic` and `Vec` do what's expected.
I'd also like to update the original PR message because there's a few tiny bits that are outdated.

Caellian:
Tested it, it seems to work so it's green flag from me as well.

I added a `From<*>` implementation for all `Value::*` variants so that `MultiValue::push_front` can now take any `Into<Value>` which makes it less verbose to use.

The implementation with a counter reference was a bit simpler to implement (less code), but I see how it could've been confusing for people.

Sorry it took me this much time, I had a lot of colloquiums and seminars at the uni I had to prepare for all at once and work.

